PR #151, opened 2016-02-27
Identity management RFC

deontologician:
This branch is an experiment. Usually when we have large features, they get discussed a lot and changed by different comments in the thread. It's often difficult to see exactly what is being proposed currently.

Discussion of the feature should happen in this pull request, and changes to the proposal should just be commits to this branch. Once the pull request is merged, there's a final version of how the feature should be implemented, hopefully unambiguously.

The issue for identity management is #3.

[Rendered markdown of the RFC](https://github.com/rethinkdb/horizon/blob/rfc/identity_mgmt/rfcs/identity_mgmt.md)

### DIFF
--- a/rfcs/identity_mgmt.md
+++ b/rfcs/identity_mgmt.md
@@ -13,6 +13,8 @@ users depending on their needs.
 
 ### Proposed solution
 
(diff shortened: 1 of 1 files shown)


deontologician:
This was completed

PR #286, opened 2016-04-23
move connection and auth constants to separate file

flipace:
This PR moves constants to a separate file so they can be easily used anywhere and not only in one place.

---
This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/286)


deontologician:
Sorry, this fell through the cracks again. We've had so much churn it's broken again :/

Looking at the code itself, I think it'd be better to store just the strings in the constants file, rather than the `{type: CONSTANT}` objects. That way they're more useful across files

deontologician:
Ok, I think I'll close this for now, since it's so bitrotted. A large refactoring is coming in #531 so this will be even more behind. Thanks @flipace for the PR

Issue #105, opened 2016-02-09
Composition of queries


deontologician:
From #93 

1. The client should be able to compose separate queries into a single object that describes the model the app needs. `fusion.model({foo: fusion('foo').find({id: myFoo}), bar: fusion('bar').find({id: myBar})})`
2. The server should be able to coalesce a composed query like this into a single union changefeed. This would allow efficient updates and the possibility of doing incremental updates of a client-side model when any constituent part of the model changes, without opening lots of separate changefeeds.

The idea is that right now, the fusion client has no idea what kind of model you're trying to build. Each query/subscription is a separate entity and the server needs to manage multiple changefeeds to the database because it doesn't know that they'll be created or destroyed together. In practice, this flexibility is needed less with React style apps (I'm including Cycle.js and Elm in this as well). In these apps, one big model is created all at once to represent the app state.

deontologician:
Implemented in #329

PR #329, opened 2016-05-06
Implement aggregated queries

deontologician:
This will implement #105 
Todo:
- [X] Code it up
- [X] Tests for `horizon.aggregate(x).fetch()`
- [x] Tests for `horizon.aggregate(x).watch()`
- [x] Tests for `horizon.model(x).fetch()`
- [x] Tests for `horizon.model(x).watch()`

---
This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/329)


deontologician:
Going to try to squeeze this into 2.0

client/src/model.js
```diff
@@ -0,0 +1,169 @@
...
+    checkWatchArgs(watchArgs)
+    const qs = this._subqueries.map(x => x.watch())
+    if (qs.length === 0) {
+      return Observable.empty()
+    } else if (qs.length === 1) {
+      return qs[0]
+    } else {
+      const headQuery = qs[0]
+      const tailQueries = qs.slice(1)
+      return headQuery.combineLatest(...tailQueries, (...args) => {
```
marshall007: `Observable.combineLatest(...qs, /* ... */)`

client/src/model.js
```diff
@@ -0,0 +1,169 @@
...
+
+  fetch() {
+    const observs = this._aggregateKeys.map(([ k, term ]) => {
+      // We jam the key into the observable so when it emits we know
+      // where to put it in the object
+      return term.fetch().map(val => [ k, val ])
+    })
+    return Observable.forkJoin(...observs, (...keyVals) => {
+      // reconstruct the object
+      const finalObject = {}
```
marshall007, line 124: It may be worth noting that `AggregateTerm` is just a special case of `ArrayTerm`. It's possible to combine both behaviors into a single class simply by changing this line to:

``` js
const finalObject = Array.isArray(aggregateObject) ? [] : {}
```

Might be worth being explicit here, I'm just not sure.

client/src/model.js
deontologician: crap, how did I miss that :+1:

client/src/model.js
deontologician, line 124: I will check it out

deontologician, line 124: Ok, after checking it out, while you're right this could be DRY-er, I think it obfuscates the code a bit. Right now there's a class with the responsibility of handling arrays and a class for handling objects, and their internal similarity is sort of incidental, not necessary.

deontologician, line 124: I also doubt this will violate the rule of 3 any time soon. I expect any further classes along these lines to be either Array or Object subtypes, not some mixture that requires we abstract out common implementation

marshall007, line 124: Yep, makes sense.

client/src/util/is-recursively-primitive.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+  if ([ 'boolean', 'number', 'string' ].indexOf(typeof value) !== -1) {
+    return true
+  }
+  if (value instanceof Date || value instanceof ArrayBuffer) {
+    return true
+  }
+  const isPOJO = isPlainObject(value)
+  const isArray = Array.isArray(value)
+  if (!isPOJO && !isArray && typeof value === 'object') {
```
marshall007, line 25: We should maybe avoid recursion here and just make this function `isPrimitive`. The `ArrayTerm` and `AggregateTerm` constructors already recurse through and `aggregate(...)` their arguments, meaning we're re-running this logic at every nesting level for every value.

deontologician, line 25: Ok, I wasn't sure you were right, but looking at it, it seems so. :+1:

client/src/model.js
```diff
@@ -0,0 +1,164 @@
...
+    return this._value.toString()
+  }
+
+  fetch() {
+    return Observable.of(this._value)
+  }
+
+  watch(...watchArgs) {
+    checkWatchArgs(watchArgs)
+    return Observable.of(this._value)
```
marshall007, line 43: @deontologician similar to our discussion earlier concerning `Array/AggregateTerm`, a separate `ObservableTerm` is not strictly necessary. Since `Observable.from(<IObservable>)` is a noop, we could get away with just having a `ValueTerm` to handle everything else (primitives, observables, etc).

I think this one is a bit more clear-cut since all we're really doing here is wrapping `of` / `from`. I think it's unlikely we would diverge too far from that in terms of what we allow you to cast from.

Also, looks like the current implementation doesn't support promises, but by using `Observable.from(...)` here we'd get that for free.

client/src/util/is-recursively-primitive.js
marshall007, line 25: Yea only difference is we'd now return `ArrayTerm<PrimitiveTerm>` instead of `PrimitiveTerm`, which seems fine if not expected.

PR #353, opened 2016-05-13
Refactor auth to use promises, expose user profile in handshake

marshall007:
Exposes the user's profile to the client in the handshake response (which will be necessary for #3 and/or #151). Other than that, this is mostly just cleaning up some nested callbacks in favor of Promises.


marshall007:
Closing in favor of #419.

PR #395, opened 2016-05-17
Pass on errors during metadata initialization

### LINKED ISSUE #392: Websocket error after upgrading to 1.0
Hello! Today I updated horizon to 1.0 and the client could not connect anymore to the server via websocket.

Running horizon with `hz serve --dev` the only response that I'm getting from the websocket is this:

```
{
  code: 1,
  message: "Session ID unknown"
}
```

Then I tried to run the server with the `--debug` flag, and that's the output:

```
debug: Handling HTTP request to horizon subpath: horizon.js
debug: Client connection established.
debug: Closing client connection with message: No connection to the database.
debug: info: {"error":"No connection to the database."}
debug: Client connection terminated.
debug: Handling HTTP request to horizon subpath: horizon.js.map
```

I also tried to create a new app with `hz init` but without luck.

As I said I'm running horizon 1.0 on node 5.10, the OS is Windows 7 x64.

danielmewes:
This seems to fix https://github.com/rethinkdb/horizon/issues/392

I'm not sure why this was working before the latest `admin` user change.

### DIFF
--- a/server/src/metadata.js
+++ b/server/src/metadata.js
@@ -231,8 +231,9 @@ class Metadata {
       // Redirect the 'users' table to the one in the internal db
       this._collections.set('users', new Collection({ id: 'users', table: 'users' },
                                                     this._internal_db, this._conn));
-    }).catch(() => {
+    }).catch((e) => {
       this.close();
+      throw e;
     });
   }
 


server/src/metadata.js
```diff
@@ -231,8 +231,9 @@ class Metadata {
       // Redirect the 'users' table to the one in the internal db
       this._collections.set('users', new Collection({ id: 'users', table: 'users' },
                                                     this._internal_db, this._conn));
-    }).catch(() => {
+    }).catch((e) => {
       this.close();
+      throw e;
```
danielmewes, line 236: If we just swallow the error during metadata initialization, we never retry. But we need to retry until we succeed (if the server is still starting up).

Tryneus, line 236: I think this should just chain onto `this._ready_promise` rather than replacing it.

Like:

```
    [...]
    }).then(() => {
      [...]
    })

    this._ready_promise.catch(() => {
      this.close();
    });
```

danielmewes, line 236: Oh that's another option. Do you think that would be better?

Tryneus, line 236: I'll upload a new PR in a minute - there were a couple other wonky things I noticed.

danielmewes, line 236: Ok. Reverting this for now.

PR #397, opened 2016-05-17
fixing readiness in server metadata syncing

Tryneus:
This does an alternate solution for #395, but also includes moving the `_ready` variable setting to true until we're actually fulfilling the promise - so nothing things we're good to go before we are.

### DIFF
--- a/server/src/metadata.js
+++ b/server/src/metadata.js
@@ -234,7 +234,7 @@ class Metadata {
       logger.debug('metadata sync complete');
       this._ready = true;
       return this;
-    })
+    });
 
     this._ready_promise.catch(() => {
       this.close();
--- a/server/src/metadata.js
+++ b/server/src/metadata.js
@@ -172,12 +172,8 @@ class Metadata {
                 }
               }).catch(reject);
             });
-          });
-      return Promise.all([ groups_ready, collections_ready, indexes_ready ]).then(() => {
-        logger.debug('metadata sync complete');
-        this._ready = true;
-        return this;
-      });
+          })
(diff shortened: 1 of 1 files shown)


server/src/metadata.js
```diff
@@ -188,6 +184,7 @@ class Metadata {
         r.expr([ this._db, this._internal_db ])
          .concatMap((db) => r.branch(r.dbList().contains(db), [], [db]))
          .run(this._conn).then((missing_dbs) => {
+           console.log("checking for internal db/tables");
```
Tryneus: this should be `logger.debug`

danielmewes: I assume this shouldn't be here

dalanmiller: => logger.debug

server/src/metadata.js
```diff
@@ -228,10 +226,17 @@ class Metadata {
...
+      logger.debug('redirecting users table');
       // Redirect the 'users' table to the one in the internal db
       this._collections.set('users', new Collection({ id: 'users', table: 'users' },
                                                     this._internal_db, this._conn));
-    }).catch((e) => {
+    }).then(() => {
+      logger.debug('metadata sync complete');
+      this._ready = true;
+      return this;
+    })
```
danielmewes: Is this missing a semicolon?

danielmewes: ... well it seems to work. Is it implicit?

Tryneus: Fixed

Tryneus: Yes, the semicolons are optional - we actually have a bunch of eslint style complaints after last night's adventures that I'll get to soon, but they're not too important.

PR #416, opened 2016-05-18
minor change to make tabs consistent

NathanielInman:
A great example, only made minor changes to tabs so that they would be consistent throughout the file.


danielmewes:
Looks good to me. Thanks for submitting a PR @NathanielInman !

PR #419, opened 2016-05-18
Refactor auth to use promises, expose user profile in handshake

marshall007:
Supersedes #353. Rebased on top of the 1.0 updates, taking into account @Tryneus's feedback on the previous PR.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/419)

### DIFF
--- a/server/src/auth.js
+++ b/server/src/auth.js
@@ -83,8 +83,6 @@ class Auth {
         .do(auth_user => insert('users', this.new_user_row(auth_user('user_id'))));
     }
 
-    query = query.merge({ provider })
-
     return this.reql_call(query)
     .catch(err => {
       // TODO: if we got a `Duplicate primary key` error, it was likely a race condition
@@ -92,7 +90,7 @@ class Auth {
       logger.debug(`Failed user lookup or creation: ${err}`);
       throw new Error('User lookup or creation in database failed.');
     })
-    .then(user => this._jwt.sign(user));
+    .then(user => this._jwt.sign({ id: user.id, provider }));
   }
 
   reql_call(query) {
@@ -119,7 +117,7 @@ class JWT {
   }
 
   // A generated token contains the data:
-  // { id: <uuid>, groups: [ ... ], provider: <string> }
+  // { id: <uuid>, provider: <string> }
   sign(payload) {
     const token = jwt.sign(
       payload,
--- a/server/src/auth.js
+++ b/server/src/auth.js
@@ -113,13 +113,13 @@ class JWT {
       this.secret = new Buffer(options.token_secret, 'base64');
     } else {
       throw new Error(
-        'No token_secret set! Try setting it in .hz/config.toml' +
+        'No token_secret set! Try setting it in .hz/config.toml ' +
         'or passing it to the Server constructor.');
     }
   }
 
   // A generated token contains the data:
-  // { user: <uuid>, provider: <string> }
+  // { id: <uuid>, groups: [ ... ], provider: <string> }
   sign(payload) {
     const token = jwt.sign(
       payload,
--- a/server/src/auth.js
+++ b/server/src/auth.js
@@ -5,50 +5,54 @@ const options_schema = require('./schema/server_options').auth;
 const writes = require('./endpoint/writes');
 
 const Joi = require('joi');
-const jwt = require('jsonwebtoken');
+const Promise = require('bluebird');
+const jwt = Promise.promisifyAll(require('jsonwebtoken'));
 const r = require('rethinkdb');
 const url = require('url');
 
-// Can't use objects in primary keys, so convert those to JSON in the db (deterministically)
-const auth_key = (provider, info) => {
-  if (info === null || Array.isArray(info) || typeof info !== 'object') {
-    return [ provider, info ];
-  } else {
-    return [ provider, r.expr(info).toJSON() ];
-  }
-};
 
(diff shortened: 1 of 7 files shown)


server/src/client.js
```diff
@@ -101,72 +101,16 @@ class Client {
       return this.close({ error: 'Invalid handshake.', error_code: 0 });
     }
 
-    const success = (user_feed, token) => {
```
marshall007, line 104: @Tryneus wasn't sure about the `user_feed` stuff here and following and why we need it.

server/src/auth.js
```diff
@@ -60,97 +64,77 @@ class Auth {
...
+      query = insert('users_auth', { id: key, user_id: r.uuid() })
+        .do(auth_user => insert('users', this.new_user_row(auth_user('user_id'))));
     }
 
-    this.reql_call(query).then((res) => {
-      cb(null, this._jwt_from_user(res.id, provider));
-    }).catch((err) => {
+    query = query.merge({ provider })
+
+    return this.reql_call(query)
```
marshall007, line 86: Note the `query` here returns the relevant row from the `users` table (regardless of which branch above is hit). We then merge in the `provider`, and use that object to sign the JWT.

server/src/auth/twitter.js
```diff
@@ -121,7 +121,7 @@ function twitter(horizon, raw_options) {
                 logger.error(`Bad JSON data from oauth API: ${body}`);
                 auth_utils.do_redirect(res, make_failure_url('unparseable inspect response'));
               } else {
-                horizon._auth.generate_jwt(provider, user_id, (err3, jwt) => {
+                horizon._auth.generate(provider, user_id).nodeify(err3, jwt) => {
```
Tryneus: This is missing a `(` before `err3`.

server/src/auth/utils.js
```diff
@@ -172,7 +172,7 @@ const oauth2 = (raw_options) => {
                 res.statusCode = 500;
                 res.end('unparseable inspect response');
               } else {
-                horizon._auth.generate_jwt(provider, user_id, (err3, jwt) => {
+                horizon._auth.generate(provider, user_id).nodeify(err3, jwt) => {
```
Tryneus: This is missing a `(` before `err3`.

server/src/auth.js
```diff
@@ -60,97 +64,77 @@ class Auth {
...
+  constructor(options) {
+    this.duration = options.duration;
+    this.algorithm = 'HS512';
+
+    if (options.token_secret != null) {
+      this.secret = new Buffer(options.token_secret, 'base64');
     } else {
-      this.verify_jwt(this._jwt_from_user(null, null), cb);
+      throw new Error(
+        'No token_secret set! Try setting it in .hz/config.toml' +
```
Tryneus: Needs a space after `config.toml`

server/src/client.js
Tryneus, line 104: The `user_feed` is used for permissions so that if a user's groups change in the database, we make sure that this session only uses rules from the new set of groups.  In addition, this updates the `Ruleset`s for all active requests on this session so they will error if they are no longer allowed.

server/src/auth/twitter.js
Tryneus: Also `make_success_url` should be passed `jwt.token`, since it is no longer a string, but an object `{ token, payload }`.  This applies to `auth/utils.js` as well.

server/src/auth.js
Tryneus, line 86: We shouldn't be sending back the entire user row, just the ID.  The client doesn't need that information during the handshake, and the auth providers don't need that information during the OAuth flow.  The `Client` class could use that information, but it needs to construct a changefeed on the user row anyway, so it'll get the data form there.

Including more data makes the JWT larger, and a user with sufficiently many group memberships could result in a query string so long that some browsers wouldn't handle it (judging by a rudimentary googling).  If the client needs this data, they can just `fetch` or `watch` the `currentUser()` method.

marshall007:
@Tryneus this should be ready for another look when you have a minute.

Tryneus:
Looks good to me - there are a few eslint violations and some test failures due to the handshake, but I'll clean it up from `next`.  Thanks for the help, @marshall007!

deontologician:
This broke the client tests. It's not flattening the payload stuff

marshall007:
@deontologician I think previously the handshake response looked like this:

```
{
  "request_id": <NUMBER>,
  "token": <STRING>,
  "user_id": <STRING>
}
```

Now you get back the `token` and then the entire decoded JWT in `payload`:

```
{
  "request_id": <NUMBER>,
  "token": <STRING>,
  "payload": {
    "id": <STRING>,
    "provider": <STRING>,
    "exp": <NUMBER>,
    "iat": <NUMBER>
  }
}
```

The intention was to avoid collisions between fields in the JWT payload and the handshake response in case either changes. I thought we would end up with more, potentially arbitrary, stuff in `payload`, but we decided to keep it simple. I can flatten the `payload` back into the response if you think that's the way to go.

marshall007:
@deontologician oh also, `payload.provider` is now set to `unauthenticated` or `anonymous` if you're using one of those auth methods. Previously using those methods just signed the token with `provider: null`.

deontologician:
@Tryneus fixed this on another branch, so we're good to go

Issue #411, opened 2016-05-18
Global install of horizon has giant test database files


josephg:
Just installed horizon. These giant files are now on my computer:

```
$ du -h /usr/local/lib/node_modules/horizon/node_modules/\@horizon/server/
 74M    /usr/local/lib/node_modules/horizon/node_modules/@horizon/server//rethinkdb_data
 48M    /usr/local/lib/node_modules/horizon/node_modules/@horizon/server//rethinkdb_data_test
```

Not sure if these files are added during post-install or are part of the module, but they shouldn't be there.

```
$ grep version /usr/local/lib/node_modules/horizon/package.json 
  "version": "1.0.1"
```

danielmewes:
Thanks for bringing this to our attention @josephg. Those files indeed shouldn't be there. We must have accidentally packaged them into the NPM.

Pinging @dalanmiller .

dv336699:
@jsehpg out of curiosity, do you always inspect what you global install?

deontologician:
I am really glad he did, dang!

josephg:
@diego-vieira only recently after I noticed another package I was using pulled in 250 megs of dependancies via npm. "I wonder how big horizon is? Whoa how is this 150 megs? ... Oh, look at that"

dalanmiller:
I'm late to this party but I take full blame for this @josephg as I was the perpetrator here, I apologize profusely to your hard drive 😭 . 

Closing this as we have proper `.gitignore` rules now.

josephg:
^_^ bugs will come and go. But seeing everyone rally to get all these issues fixed makes me really happy. Thanks for the great work!

PR #430, opened 2016-05-19
Add rethinkdb_data to gitignore

deontologician:
Part of #411 

@dalanmiller could you check this out and decide if we need to do anything else to prevent data from getting into the npm package?

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/430)

### DIFF
--- a/client/.gitignore
+++ b/client/.gitignore
@@ -28,3 +28,6 @@ build/
 # https://docs.npmjs.com/misc/faq#should-i-check-my-node-modules-folder-into-git
 node_modules
 
+# database stuff
+rethinkdb_data/
+rethinkdb_data_test/
--- a/server/.gitignore
+++ b/server/.gitignore
@@ -5,3 +5,7 @@ node_modules
 
 # Coverage directory used by tools like istanbul
 coverage
+
+# RethinkDB stuff
+rethinkdb_data/
+rethinkdb_data_test/


danielmewes:
Wonder if this is really the right approach, since there were other files (rethinkdb_data_test) included in the current NPM as well.

It is also possible to configure _positive_ rules by negating a line in .npmignore. We could limit to certain file endings. Would that make sense?
https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package

danielmewes:
Actually as far as I could find out, we'd need to have a set of file ending rules for each directory depth. It becomes a bit annoying. (Edit: or maybe not, `**/` might work for this)

Maybe another option is writing our own short `publish.sh` script, that does the following:
1. Call `git status | grep -q "working directory clean"` and check the return code. If it returns anything other than 0, we print an error and abort.
2. Call `npm publish`

danielmewes:
Such a script might be better since it generally stops us from publishing uncommitted changes or any extra files.

deontologician:
That would be pretty nice.

danielmewes:
@dalanmiller How exactly and from where do you run `npm publish`? I'm happy to write the script.

deontologician:
I opened #431 to track creating such a script. I still think this PR is useful since we shouldn't accidentally commit rethinkdb_data directories to git either

deontologician:
Reviewed 2 of 2 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/430)_

PR #439, opened 2016-05-19
Remove auth_redirect from yes_no_options

mlynch:
auth_redirect in config.toml takes a string argument specifying the URL to redirect the user to for auth. It was incorrectly being enforced as a boolean so it could not be configured in the config file.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/439)


deontologician:
Seems good. ping @Tryneus

PR #440, opened 2016-05-19
--permissions flag missing from cli readme

BrianRosamilia:
This tripped me up when running hz serve with a docker rethinkdb

I couldn't figure out how to disable the permissions system until I found this page https://github.com/rethinkdb/horizon#readme and learned of the --permissions flag

That page also mentions that --dev will set the permissions flag.  This is also omitted from the readme.

This PR fixes both documentation issues.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/440)


deontologician:
Thanks! Sorry for the confusion

PR #453, opened 2016-05-19
fix eslint violations in the server and cli

Tryneus:
There is at least one crash fix here - `path` was not being required in `server/server.js`

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/453)


deontologician:
:lgtm:

Issue #376, opened 2016-05-17
`hz init .` generates a config which sets `project_name` to `.`, which isn't a legal database name.


mlucy:
Probably we should make this work and set the `project_name` to the current directory.

In general we should probably either be validating or escaping project names.

oliver006:
Similar problem: 
I'm following the tutorial instructions and create an app via `hz init`. If I use an app name that's illegal for a rethinkdb database then the subsequent `hz serve` fails. Example would be: `hz init hz-app`

cletusw:
The [Getting Started guide](http://horizon.io/docs/getting-started/#initialize-an-example-application) itself recommends using `hz init example-app`, which fails on the `hz serve` step for this same reason.

```
$ hz init example-app
$ cd example-app/
$ hz serve --dev
App available at http://127.0.0.1:8181
RethinkDB
   ├── Admin interface: http://localhost:53716
   └── Drivers can connect to port 53715
Starting Horizon...
error: Connection to RethinkDB terminated: ReqlQueryLogicError: Database name `example-app` invalid (Use A-Za-z0-9_ only) in:
r(["example-app", "example-app_internal"]).forEach(function(var_0) { return r.branch(r.dbList().contains(var_0), [], r.dbCreate(var_0)); }).do(function() { return r(["collections", "users_auth", "users", "groups"]).forEach(function(var_1) { return r.branch(r.db("example-app_internal").tableList().contains(var_1), [], r.db("example-app_internal").tableCreate(var_1)); }); })
```

`hz init` should run a regex and replace anything that doesn't match `A-Za-z0-9_` with `_`

deontologician:
Yeah, this is definitely a wart

danielcompton:
Might pay to update the getting started docs to recommend `example_app` until this can be fixed.

stewartellis:
Glad this has been reported. I just installed the repo for the first time tonight and thought I had some kind of funky RethinkDB config issue.

mglukhovsky:
Getting started docs have been updated to `example_app`.

danielmewes:
I think the best behavior for `hz init` would be to:
- Replace `-` by `_` in the project name before writing it to the config file
- If the path contains any characters other than `[A-Za-z0-9_-]`, print and error and abort creating the project. This handles the case where you do `hz init .`, so you don't end up with a project called `_`. It also helps if for example you're trying to use let's say Chinese characters with `hz init`. Simply rewriting the project name to `________` would be bad.

[...]

PR #472, opened 2016-05-21
Cli cwd changes

deontologician:
Implements #376 

```
 checkProjectName
    when passed null for a directory
      ✓ doesn't change directory
      ✓ throws if the cwd has invalid chars
      ✓ sets projectName to dehyphenated cwd if fixable
    when passed "." as a directory
      ✓ doesn't change directory
      ✓ throws if the cwd has invalid chars
      ✓ sets projectName to dehyphenated cwd if fixable
    when passed a non-existing directory
      ✓ creates the directory when name is valid
      ✓ creates the directory when name is fixable
      ✓ throws an error if the name is not fixable
    when passed an existing directory
      ✓ errors if given an invalid projectName
      ✓ changes directory  if the name is good
      ✓ changes directory if the name is fixable
```

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/472)

### DIFF
--- a/cli/src/main.js
+++ b/cli/src/main.js
@@ -14,115 +14,59 @@ const getSchemaCommand = require('./get-schema');
 const setSchemaCommand = require('./set-schema');
 const makeTokenCommand = require('./make-token');
 
-const parser = new argparse.ArgumentParser();
-
-const subparsers = parser.addSubparsers({
-  title: 'commands',
-  dest: 'command_name',
-});
-
-const initParser = subparsers.addParser('init', {
-  addHelp: true,
-  help: 'Initialize a horizon app directory',
-});
-
-const serveParser = subparsers.addParser('serve', {
-  addHelp: true,
-  help: 'Serve a horizon app',
-});
-
-const versionParser = subparsers.addParser('version', { // eslint-disable-line no-unused-vars
-  addHelp: true,
-  help: 'Print the verison number of horizon',
-});
-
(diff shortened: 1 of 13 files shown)


marshall007:
One minor thing on this, I think you should be able to specify an "invalid" directory name (containing hyphens) and we should only worry about sanitizing the project name in the `config.toml`.

deontologician:
Good point

deontologician:
Ok, this is ready to review. Any takers @Tryneus or @dalanmiller ?

cli/src/main.js
```diff
@@ -72,36 +72,55 @@ const done_cb = (options) => (err) => {
   }
 };
 
+function runCommand(command, options, done) {
```
Tryneus: since `done` is always `done_cb(options)`, that parameter is extraneous.

deontologician:
I went ahead and made the refactors to cli main.js, since I was feeling less lazy today

cli/src/main.js
```diff
@@ -14,96 +14,59 @@ const getSchemaCommand = require('./get-schema');
...
-  help: 'Get the schema from a horizon database',
-});
-
-const setSchemaParser = subparsers.addParser('set-schema', {
-  addHelp: true,
-  help: 'Set the schema in a horizon database',
-});
+// Mapping from command line strings to modules. To add a new command,
+// add an entry in this object, and create a module with the following
+// exported:
```
dalanmiller: "following functions exported"?

dalanmiller:
:lgtm:

PR #473, opened 2016-05-21
Fix changefeed handling and make-token

Tryneus:
Three major changes:
- `make-token` was generating tokens with the old format, using 'user' instead of 'id'
- changefeeds from `query`, `subscribe`, and `get_user_feed` were not being `catch`ed when closed
- the `user_feed` in `Client` was not being handled properly

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/473)

### DIFF
--- a/cli/src/make-token.js
+++ b/cli/src/make-token.js
@@ -105,7 +105,7 @@ const runCommand = (options, done) => {
       throw new Error('User does not exist.');
     }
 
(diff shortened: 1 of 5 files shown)


danielmewes:
Reviewed 5 of 5 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/473)_

PR #474, opened 2016-05-21
Make tests pass on CircleCI

danielmewes:
@dalanmiller

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/474)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -57,14 +57,14 @@ test:
     - touch $CIRCLE_TEST_REPORTS/xunit/server-tests.xml
   override:
     # Run client tests
-    - ./node_modules/.bin/mocha --timeout 30000 -R xunit dist/test.js > $CIRCLE_TEST_REPORTS/xunit/client-tests.xml:
+    - ./node_modules/.bin/mocha --timeout 90000 -R xunit dist/test.js > $CIRCLE_TEST_REPORTS/xunit/client-tests.xml:
         pwd: client
-        parallel: true
+        parallel: false
     # Run server tests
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -57,14 +57,14 @@ test:
     - touch $CIRCLE_TEST_REPORTS/xunit/server-tests.xml
   override:
     # Run client tests
-    - ./node_modules/.bin/mocha --timeout 30000 -R xunit dist/test.js > $CIRCLE_TEST_REPORTS/xunit/client-tests.xml:
```
danielmewes, line 60: I bumped the timeouts up further. Because why be stingy... Also the server tests sometimes took slightly over 30 seconds.

circle.yml
```diff
@@ -57,14 +57,14 @@ test:
...
+    - ./node_modules/.bin/mocha --timeout 90000 -R xunit dist/test.js > $CIRCLE_TEST_REPORTS/xunit/client-tests.xml:
         pwd: client
-        parallel: true
+        parallel: false
```
danielmewes, line 62: I don't understand how CircleCI's parallelism works exactly, but it seemed to break the tests. Almost as if some of the `pre` steps weren't properly applied across all parallel instances.

server/test/http_tests.js
```diff
@@ -61,7 +61,7 @@ const all_tests = () => {
       });
 
       it('localhost/horizon/horizon.js', (done) => {
-        require(transport).get({ host: http_server.address().address,
+        require(transport).get({ host: "localhost",
```
danielmewes, line 64: It looks like `address` isn't resolving properly on CircleCI. I think it should be ok to always use `localhost` here.

dalanmiller:
Merged at 474a28c2fb4c7c8985efd0f1bbb0d528fc798c72

PR #480, opened 2016-05-22
Set process.title properly

steelbrain:
To support `pidof horizon`, by default it shows in `pidof node`
This will save us from accidental kill when we are trying to kill another node app instead of horizon by `killall node`

PS. I've signed the CLA

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/480)


deontologician:
looks good to me, thanks!

PR #483, opened 2016-05-22
remove redundant and incorrect horizon.js load

Beasta:
This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/483)


dalanmiller:
Thank you @Beasta! That was an artifact of ensuring the client library was loaded for examples when using https or http. I think using the relative path will work pretty much most of the time as it follows the document's transport type.

PR #489, opened 2016-05-23
update horizon command line paramter

moklick:
I tried `--insecure` which doesn't exist. `--secure no` works

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/489)

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -53,7 +53,7 @@ npm run devtest     | Run tests and linting continually
 
 Object which initializes the connection to a Horizon Server.
 
-If Horizon server has been started with `--insecure` then you will need to connect unsecurely by passing `{secure: false}` as a second parameter.
+If Horizon server has been started with `--secure no` then you will need to connect unsecurely by passing `{secure: false}` as a second parameter.
 
 ###### Example
 


danielmewes:
Thanks for the PR @moklick . The change looks good to me.

I wonder though if the sentence is still true. At least when connecting locally, connecting seems to work fine to an insecure server even without passing `{secure: false}` to the `Horizon` constructor. @deontologician do you know what the current behavior is in this regard?

We should also consider replacing the README by a link to the Horizon API documentation in https://github.com/rethinkdb/horizon-docs/tree/master/api , so we don't have to maintain two copies of it.

deontologician:
:lgtm: I'll open an issue for changing the readme

<details><summary>

<i>Previously, danielmewes (Daniel Mewes) wrote…</i></summary>



> Thanks for the PR @moklick . The change looks good to me.
> 
> I wonder though if the sentence is still true. At least when connecting locally, connecting seems to work fine to an insecure server even without passing `{secure: false}` to the `Horizon` constructor. @deontologician do you know what the current behavior is in this regard?
> 
> We should also consider replacing the README by a link to the Horizon API documentation in https://github.com/rethinkdb/horizon-docs/tree/master/api , so we don't have to maintain two copies of it.
> </details>

---

Reviewed 1 of 1 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/489#-:-KIUjYVhkrWImejifU-7:-1112230622)_

PR #490, opened 2016-05-23
Fixes to make more consistently passing tests.

dalanmiller:
Fixes to make more consistently passing tests. Also, allows tests to work even when the current upped version hasn't been pushed to npm yet. 

@tryneus, does the hardset timeout in the server tests need to be present?

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/490)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -57,14 +57,14 @@ test:
     - touch $CIRCLE_TEST_REPORTS/xunit/server-tests.xml
   override:
     # Run client tests
-    - ./node_modules/.bin/mocha --timeout 90000 -R xunit dist/test.js > $CIRCLE_TEST_REPORTS/xunit/client-tests.xml:
+    - ./node_modules/.bin/mocha --timeout 100000 -R xunit dist/test.js > $CIRCLE_TEST_REPORTS/xunit/client-tests.xml:
         pwd: client
         parallel: false
     # Run server tests
-    - ./node_modules/.bin/mocha --timeout 90000 -R xunit test/test.js test/schema.js > $CIRCLE_TEST_REPORTS/xunit/server-tests.xml:
+    - ./node_modules/.bin/mocha --timeout 100000 -R xunit test/test.js test/schema.js > $CIRCLE_TEST_REPORTS/xunit/server-tests.xml:
         pwd: server
         parallel: false
     # Run cli tests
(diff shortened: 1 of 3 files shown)


danielmewes:
@dalanmiller I think I added the `this.timeout(30000);` to the server test. The reason is that Mocha has a default timeout which was way too low (I think it's like 2 seconds). Note that this timeout is _per test_. I think it's possible to disable per-test timeouts alltogether, but it has the disadvantage that a single stuck tests will cause the overall server test suite to time out rather than timing out individually.

dalanmiller:
@danielmewes was `mocha` not accepting the timeout args from the CLI flag?

danielmewes:
@dalanmiller I think the command-line flag is a global timeout, across all tests? It's also a good idea to make it work without any flags on the command line I think.

deontologician:
:lgtm:

<details><summary>

<i>Previously, danielmewes (Daniel Mewes) wrote…</i></summary>



> @dalanmiller I think the command-line flag is a global timeout, across all tests? It's also a good idea to make it work without any flags on the command line I think.
> </details>

---

Reviewed 3 of 3 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/490#-:-KIUejsTYn3FxGLQyhQm:1417088721)_

dalanmiller:
@danielmewes

>  I think the command-line flag is a global timeout, across all tests?

Yeah I think that's true.
https://mochajs.org/#usage

> It's also a good idea to make it work without any flags on the command line I think.

Technically it does 😁 :

```
npm run test
```

Given the two evils of having an arbitrarily large timeout in the command-line flag or hidden somewhere in the tests themselves. I vote going with the command-line flag.

danielmewes:
Ok I think I misunderstood what the --timeout option did. It looks like it's indeed equivalent to the `this.timeout`call inside the test:

> -t, --timeout <ms>                      set test-case timeout in milliseconds [2000]

So it's also per test-case. Getting rid of one of them sounds good then.

deontologician:
Reviewed 1 of 1 files at r2.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/490)_

Issue #493, opened 2016-05-23
Modify readme to point to horizon docs


deontologician:
We have rapidly changing information reflected in two places, so it'd be better if we didn't need to maintain the README's separately.

PR #503, opened 2016-05-24
Readme purge and redirecting to horizon.io

dalanmiller:
Satisfies #493

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/503)


deontologician:
I like this, still gives you an outline, but reduces the number of things that can get out of sync by a lot.

:lgtm:

PR #504, opened 2016-05-24
Return of the --open flag 

dalanmiller:
Re-added this since serve.js had deviated a bit from previous. Now works only with `--open` flag as desired from comments before. Only wish would be that in the case of multiple `--open`ings that it would just refresh the already opened window and tab.

https://stackoverflow.com/questions/8454510/open-url-in-same-window-and-in-same-tab

closes #223

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/504)

### DIFF
--- a/cli/src/serve.js
+++ b/cli/src/serve.js
@@ -580,23 +580,18 @@ const runCommand = (opts, done) => {
     // Automatically open up index.html in the `dist` directory only if
     //  `--open` flag specified and an index.html exists in the directory.
     if (opts.open && opts.serve_static) {
-      // Check if index.html exists and readable in serve static_static directory
-      fs.access(`${opts.serve_static}/index.html`, fs.R_OK | fs.F_OK, (err) => {
-        if (err) {
-          console.log('Could not find index.html, will not auto open');
-        } else {
-          console.log('Attempting open of index.html in default browser');
-          // Determine scheme from options
-          const scheme = opts.secure ? 'https://' : 'http://';
-          // Open up index.html in default browser
-          try {
-            open(`${scheme}${opts.bind}:${opts.port}/index.html`);
-          } catch (open_err) {
-            console.log(`Error occurred while trying to open ${opts.serve_static}/index.html`);
-            console.log(open_err);
-          }
-        }
-      });
+      try {
+        // Check if index.html exists and readable in serve static_static directory
+        fs.accessSync(`${opts.serve_static}/index.html`, fs.R_OK | fs.F_OK);
+        // Determine scheme from options
+        const scheme = opts.secure ? 'https://' : 'http://';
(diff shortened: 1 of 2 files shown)


deontologician:
:lgtm: I'll leave it to you to merge it, depending on whether you feel like switching to accessSync or not

<details><summary>

> #### Return of the --open flag
> 
> Re-added this since serve.js had deviated a bit from previous. Now works only with `--open` flag as desired from comments before. Only wish would be that in the case of multiple `--open`ings that it would just refresh the already opened window and tab.
> 
> https://stackoverflow.com/questions/8454510/open-url-in-same-window-and-in-same-tab
> 
> closes #223
> 
> </details>

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

_[cli/src/serve.js, line 584 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/504#-KI_FVEvZAdveHFqgRlY:-KI_FVEvZAdveHFqgRlZ:743469302) ([raw file](https://github.com/rethinkdb/horizon/blob/000a7531e67b7033c7cf08cbe6e0bbc5ef88dfe8/cli/src/serve.js#L584)):_

> ``` JavaScript
>     if (opts.open && opts.serve_static) {
>       // Check if index.html exists and readable in serve static_static directory
>       fs.access(`${opts.serve_static}/index.html`, fs.R_OK | fs.F_OK, (err) => {
> ```

You could use accessSync here, we've been using the sync versions in the cli

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/504#-:-KI_FesIS5OWlJICTqH7:1034430199)_

cli/src/serve.js
```diff
@@ -565,6 +576,28 @@ const runCommand = (opts, done) => {
                                       Object.assign({}, { path: name }, opts.auth[name]));
       }
     }
+  }).then(() => {
+    // Automatically open up index.html in the `dist` directory only if
+    //  in dev mode and an index.html exists in the directory.
```
Tryneus: why does this mention dev mode?

dalanmiller: Originally, it opened the browser when in dev mode. I just forgot to take this comment out.

dalanmiller:
@deontologician appropriate use of `accessSync`?

https://github.com/rethinkdb/horizon/pull/504/files#diff-7e8aa0fffc19f1f930c3f20aff511503R585

deontologician:
Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/504)_

PR #506, opened 2016-05-25
POC for custom auth

stellanhaglund:
I just tried this out mainly because I need it myself, and hopefully its in some way usable! :)

The code could probably be done a little better, and I'm not sure it would make any sense allowing multiple auth modules, but it should work.
I added an option to the server so that this could be enabled or disabled, disabled wouldn't change anything from the default setup.
Right now it allows you to combine your own parts with parts from the server so I think it does it's purpose.

a small examples would be

```
server.registerAuthModule('customModule', (hz, handshake, jwt) => {

    var parsed = hz.parse_request(handshake, 'handshake') // <-- the only reason for the switch part which isnt an optimal solution

    // verify a token
    jwt.verify(parsed.token).then((res) => {

    })

  // pass a response back
  function finish_custom_handshake(data){
    var inf = { 
      id: 'undefined',
      provider: 'custom',
      token: token.token
    }
    hz.send_response(data, inf);
    hz._socket.on('message', (msg) => {
      hz.error_wrap_socket(() => {
        hz.handle_request(msg)
      })
    })
  }

  finish_custom_handshake(parsed)


})

```

I think that you understand what could be done in here.

I thought this would be good because it really doesn't change any existing logic and that it would be good the get this added now so that it follows in future relases.

Would be glad to hear what you think.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/506)

### DIFF
--- a/server/src/client.js
+++ b/server/src/client.js
@@ -24,10 +24,19 @@ class Client {
 
     this._socket.on('error', (error) =>
       this.handle_websocket_error(error));
-
+    
+    // not pretty, hardcoded modulename
+    this._custom = this._server._auth_modules.get('customModule')
+        
     // The first message should always be the handshake
-    this._socket.once('message', (data) =>
-      this.error_wrap_socket(() => this.handle_handshake(data)));
+    // if custom_auth is enabled pass the handshake to that instead
+    if(this._server._custom_auth){
+      this._socket.once('message', (data) =>
+        this.error_wrap_socket(() => this._custom(this, data, this._auth._jwt)));
+    }else{
+      this._socket.once('message', (data) =>
+        this.error_wrap_socket(() => this.handle_handshake(data)));    
+    }
 
     if (!this._metadata.is_ready()) {
       this.close({ error: 'No connection to the database.' });
@@ -62,6 +71,22 @@ class Client {
   }
 
   parse_request(data, schema) {
+
+    // not pretty right now, just to allow accessing this from the server in a simple way
+    switch (schema){
+      case 'request':
+        console.log('request')
+        schema = schemas.request
+        break;
+      case 'handshake':
+        console.log('handshake')
+        schema = schemas.handshake
+        break;
+      default:
+        break;
+    }
+    
+     
     let request;
     try {
       request = JSON.parse(data);
--- a/server/src/client.js
+++ b/server/src/client.js
@@ -30,18 +30,34 @@ class Client {
     // The first message should always be the handshake
     // if custom_auth is enabled pass the handshake to those instead
     if(this._server._custom_auth){
+      
+      this._custom.forEach((x) => {
+        x(this)
+      })
+      /*
       this._socket.once('message', (data) => {
         this.error_wrap_socket(() => {
           this._custom.forEach((x) => {
+            console.log('message passed to custom ')
             x(this, data, this._auth._jwt)
           })
         })
         
+        
+        
       })
+      
+      */
         
     }else{
-      this._socket.once('message', (data) =>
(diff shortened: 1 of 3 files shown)


Tryneus:
I like the idea of pluggable auth, and we'll definitely need it soon - a lot of people are asking about how to integrate their own or third-party auth providers.  I think this solution in particular doesn't do enough to make this work, so I'd like to propose an API, and please let me know if it works for your case.  Given this, we should be able to pull apart the existing `unauthenticated`, `anonymous`, and third-party OAuth methods into pluggable modules that would serve as a proof-of-concept.

Now, for the interface used by these modules:

``` js
class Server {
[...]
  // `name` is the name of the auth provider, which will be signed into any tokens generated
  // `handshake_cb` will be called when a horizon client connects using `name` as its `method` parameter
  // `endpoint_cb` will be called when a browser requests the '/horizon/<name>` page - may be omitted
  add_auth_provider(name, handshake_cb [, endpoint_cb]) { [...] }
[...]
}
```

A `handshake_cb` would be of the form `function handshake_cb(handshake, metadata, jwt, done)`, where:
- `handshake` is the object parsed from the JSON handshake sent by the Horizon client
  - This should contain a `method` field matching the `name` of this auth provider
  - This may also contain extra data used to identify a specific user (may be different for each endpoint)
  - Validation of the schema of this object is left up to the `handshake_cb` function
- `metadata` is a `Metadata` object (from `metadata.js`)
  - This may be used to get a ReQL connection or determine the names of the internal or data dbs
- `jwt` is the Horizon `JWT` object which can be used to verify and decode JSON Web Tokens
- `done` is a callback accepting one argument
  - This should be called when the user has been verified or an error has occurred
  - If an `Error` object is passed in, this will be forwarded to the Horizon client, and the connection closed
  - If the argument is a ReQL changefeed, we will treat the results of that changefeed as the user information, updating in real-time.
  - Otherwise, the argument should an object of user infomation with a `groups` field indicating which permissions this session is allowed

An `endpoint_cb` is an HTTP request handler, like `function endpoint_cb(request, response, jwt)`:
- `request` is an instance of `http.IncomingMessage`
- `response` is an instance of `http.ServerResponse`
- `server` is the Horizon `JWT` object, which can be used to sign tokens with the Horizon server's secret

The `endpoint_cb` should verify the user's identity, then communicate some information back to the client (in the HTTP response) that will be used during the handshake.  For the standard third-party authentication methods, this is done by saving a JWT in local storage.  `unauthenticated` and `anonymous` methods do not need an `endpoint_cb`, because they do not require existing state during the handshake.

This might need some tweaking.  The biggest missing things I see are:
- A way to facilitate a presence API through this - I assume anyone providing their own authentication may want/need to track open sessions
- If the custom auth provider does not use the internal `users` table, then `currentUser()` and the users collection will be useless to the Horizon client.

Let me know if you see any problems or have any suggestions for this.

stellanhaglund:
So just a short question, wouldn't using the unauthenticated and  anonymous still be tied to using the internal parts for groups/users/users_auth ?

I'm thinking about this from where I stand and I don't want to use the internal tables for users or those parts, I think it would be best to have an option to use whatever parts of horizon that might be helpful, so to be able to use for example sendRequest and to generate tokens from my own table using your methods and relying on your handshakes seems like perfect setup for me.

So what I mean is that adding an api to it would really be, the way I got it to work was by finding parts in the code I needed.
But I still think that you should be able to use your own tables like I do and select what parts that might be needed.

As long as the user implementing it knows what parts depends on the internal tables there should be no problem with them not using them, or would you want to keep it focused to those tables?

Issue #288, opened 2016-04-25
Allow security options in reql_connection.js


deontologician:
Right now only host/port are accepted as options. Ideally we'd accept certificate paths, user/password etc, that 2.3 allows

danielmewes:
I've been informed that Horizon currently doesn't work with non-admin users because it accesses some of the system tables. This will be resolved with RethinkDB 2.4 https://github.com/rethinkdb/rethinkdb/issues/5692 .

danielmewes:
@Tryneus it looks like we're using system tables to a) wait for indexes to become ready and b) to check the RethinkDB server version, right? I wonder if we could implement some kind of fallback so Horizon still works with non-admin users (though possibly not properly waiting for indexes, and not checking the server version) until we ship RethinkDB 2.4.

deontologician:
Fixed by #507

PR #507, opened 2016-05-25
Security options for rethinkdb

ericb:
Additions in support of #288 
- [x] Add `user/password` fields for rethinkdb connections
- [x] Add ability to use `rethinkdb://` connection uris with the `connect` config option.
- [x] Add `timeout` setting.
- ~~Add `ssl` options~~

The use of `rethinkdb://` may or may not be a welcome addition here. I was not able to find any reference to it's use with rethinkdb directly, but following traditional db connection uri's it seemed like a good idea.

As already stated in #288, this cannot be fully utilized until the rethinkdb 2.4 release or other changes to horizon are put in place temporarily. 
#### Notes
- If you specify a db in the connection uri, it will override the project name field (as they are currently tied together)
- RethinkDB connection timeout can be set as `rdb_timeout` (config.toml), `--rdb-timeout` (cli usage), and `HZ_RDB_TIMEOUT` (ENV var)
- RethinkDB user can be set as `rdb_user` (config.toml), `--rdb-user` (cli usage), and `HZ_RDB_USER` (ENV Var)
- RethinkDB password can be set as `rdb_password` (config.toml), `--rdb-password` (cli usage), and `HZ_RDB_PASSWORD` (ENV Var)

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/507)


cmnstmntmn:
did you consider using **[Let's Encrypt](https://letsencrypt.org/)** to generate certificates ?

## 

Let’s Encrypt is a new Certificate Authority: 
It’s free, automated, and open.

danielmewes:
@cmnstmntmn That seems unrelated to this change.

ericb:
@danielmewes I just realized I never actually allowed you to set the user or password via flag / env var, and in a limited capacity for the config.toml. I'm thinking i'd like to convert these options to be `db_user, db_pass, db_timeout, etc` for naming. I think that would clear any confusion between Horizon and RethinkDB specific settings. Thoughts?

danielmewes:
Let me ping @Tryneus for this. I'm not completely sure what naming schema we're using elsewhere in the Horizon server.

cli/src/serve.js
```diff
@@ -305,6 +335,8 @@ const read_config_from_file = (project_path, config_file) => {
   for (const field in file_config) {
     if (field === 'connect') {
       parse_connect(file_config.connect, config);
+    } else if (field === 'connect_timeout') {
```
Tryneus: You shouldn't have to do this if you have a `connect_timeout` field in the `default_config` object.  It would be a bit more consistent to do it that way.

Tryneus: Same on line 368 as well.

cli/src/serve.js
```diff
@@ -404,6 +439,10 @@ const read_config_from_flags = (parsed) => {
     config.bind = parsed.bind;
   }
 
+  if (parsed.connect_timeout !== null) {
```
Tryneus: This should also make sure `parsed.connect_timeout` is not `undefined` - this function is called by the `get-schema` and `set-schema` scripts which may or may not set all of these.

cli/src/serve.js
```diff
@@ -267,18 +271,44 @@ const yes_no_options = [ 'debug',
...
-    config.rdb_host = host_port[0];
-    config.rdb_port = parseInt(host_port[1]);
-    if (isNaN(config.rdb_port) || config.rdb_port < 0 || config.rdb_port > 65535) {
-      throw new Error(`Invalid port: "${host_port[1]}".`);
+  // support rethinkdb:// style connection uri strings
+  // expects rethinkdb://host:port` at a minimum but can optionally take a user:pass and db
+  // e.g. rethinkdb://user:pass@host:port/db
+  const rdb_uri = url.parse(connect);
+  if (rdb_uri.protocol === "rethinkdb:"){
+    if (rdb_uri.hostname && rdb_uri.port) {
```
Tryneus: Not exactly sure what `url.parse` does when the port is omitted, but it should be optional.  If there is no port, we should use the default rethinkdb driver port (`const default_rdb_port = 28015;`).

Tryneus:
@ericb - The config file and env variables work automatically if you add fields to the default values object.  I don't feel too strongly on the names, the ones you suggested should be fine.

I took a look at the diff and had a few comments, but overall it looks pretty good.

ericb:
@Tryneus Ah, I completely missed the fact that it picks up those options. I'll provide an update to reflect your comments. Thanks!

alf:
Would be great if this one got included soon.

danielmewes:
@ericb We'd like to include this soon. Would you like us to make the remaining fixes? Thanks again for your work on this!

ericb:
Yeah, I'll go ahead and work on finishing it off now.

ericb:
@danielmewes @Tryneus I pushed updates based on your feedback. The only other thing to note is that while you can set the database timeout, it currently doesn't have much effect because it's caught as an error and retried automatically (up to the horizon timeout of 30 seconds).

deontologician:
Looks like there are still some conflicts with next

ericb:
Should be good to go now @deontologician , sorry about that. 👍

server/src/server.js
```diff
@@ -78,7 +78,10 @@ class Server {
                                          opts.rdb_port,
                                          opts.project_name,
                                          opts.auto_create_collection,
-                                         opts.auto_create_index);
+                                         opts.auto_create_index,
+                                         opts.rdb_user || null,
+                                         opts.rdb_pass || null,
```
alf: This should be opts.rdb_password

alf:
It's not good to go yet. Please fix the above issue as well. :)

ericb:
Thanks @alf, pushed that up.

Tryneus:
Ok, squash merged into `next` in commit 4cdf47b (there was a minor conflict).  Thanks again for helping out, @ericb, a lot of people were looking forward to this!

danielmewes:
🎉 :clap:

GeoffreyPlitt:
This doesn't seem to be included in the current published version on NPM. Why is "next" the default branch for this repo? When does "next" get merged to master? When will NPM publish the "next" branch?

GeoffreyPlitt:
I found a way to run this and the command-line seems to parse the extra password/etc values correctly, but the server fails at some validation step-- is this expected?

GeoffMacbookPro:stardust-rethink-server giro$ node ./node_modules/horizon#next/cli/src/main serve
WARNING: Serving app insecurely.
App available at http://127.0.0.1:8181
Starting Horizon...
serve failed with ValidationError: {
  "auto_create_collection": false,
  "auto_create_index": false,
  "permissions": true,
  "project_name": "test123",
  "auth": {
    "token_secret": "<SNIPPED>",
    "allow_unauthenticated": false,
    "allow_anonymous": false,
    "success_redirect": "/",
    "failure_redirect": "/"
  },
  "rdb_host": "aws-us-east-1-portal.15.dblayer.com",
  "rdb_port": "11342",
  "rdb_timeout" [1]: 20,
  "rdb_password" [2]: "<SNIPPED>",
  "rdb_user" [3]: "admin",
  "access_control_allow_origin" [4]: ""
}

[1] "access_control_allow_origin" is not allowed
[2] "rdb_user" is not allowed
[3] "rdb_password" is not allowed
[4] "rdb_timeout" is not allowed

danielmewes:
@GeoffreyPlitt We're going to release Horizon 2.0 early next week and publish it on NPM. That will contain this patch.

I'm not sure why the server would not accept these options. Are you sure that the CLI is using the right version of the Horizon server, and not the one from NPM?

GeoffreyPlitt:
Sweet! I guess we'll wait for the release.

Yes, the mono-repo thing is likely related to my challenge with this, and I think references to NPM is the issue.

ubergarm:
As a developer of a multi-tenant app, this kind of security is important to have in place! Its great to see the RethinkDB / Horizon ecosystem maturing to be more inline with my expectations (e.g. supporting the canonical scheme:[//[user:password@]host[:port]][/]path etc...)

But I'm concerned that it appears Horizon is "re-implementing" its own per app users/credentials/permissions vs using the existing RethinkDB tables.

```
r.db('rethinkdb').table('users')
r.db('rethinkdb').table('permissions')
```

vs

```
r.db('myapp_internal').table('users')
r.db('myapp_internal').table('users_auth')
```

There must be pros and cons, but now there are "two sources of truth" so to speak for credentials. PostgREST takes the approach of mapping app users to database roles:

http://postgrest.com/admin/security/

Not sure this is the place to discuss, but "two sources of truth" seems to cause problems for my use case:

1) Clients want the typical web app, but also want an REST API to their data.
2) I tell clients, "just use RethinkDB client, its better than any REST API I could give you"
3) Their Horizon based web app users don't line up with their "API access" users.

Maybe I'm going at this the wrong way, but  just something to consider while its still early in the game.

Thanks!

danielmewes:
@ubergarm There are various reasons for maintaining users of your application separately from database users, but the most important one is that RethinkDB user permissions are nowhere near granular enough to implement the complex permission requirements that applications can have.

The way you would use RethinkDB users is to isolate individual Horizon applications if they share a common RethinkDB cluster. It's not intended to map to the individual users of a single Horizon application.

danielmewes:
> 1) Clients want the typical web app, but also want an REST API to their data.
> 2) I tell clients, "just use RethinkDB client, its better than any REST API I could give you"
> 3) Their Horizon based web app users don't line up with their "API access" users.

This is an interesting scenario. Unfortunately RethinkDB's ReQL is not designed for having end-users access it directly (this was one of the reasons for why we developed a new API for Horizon, rather than just exposing ReQL). I think we'll make it as easy as possible to utilize Horizon's permission system for custom endpoints in the future though. So you could add your own REST endpoints, which would rely on internal Horizon server API capabilities for performing permission checks.

ubergarm:
@danielmewes Thanks for the update! I appreciate your reasoning and clarity.

I really don't want to add my own REST endpoints! haha.. ;)

So maybe I'll wait for a Horizon Python client or see how the GraphQL server pans out to allow clients to programmatically access app data with app appropriate permissions (for things like async tasks, in lieu of REST APIs, etc.)

Thanks again though and I appreciate your and the teams effort! "Real-time" is changing the game!

levimoore:
I realize this is closed and more than likely I am missing something, but connecting to Compose.io requires you use the Self Signed SSL certificate they provide you in the dashboard which creates an x509 error in Horizon. (Obviously)

serve failed with Error: error:0B080074:x509 certificate routines:X509_check_private_key:key values mismatch

If I use the hz create-cert files the connection just times out.

Horizon: 2.0.0
Compose.io: RethinkDB 2.3.4

GeoffreyPlitt:
+1

danielmewes:
This patch doesn't contain an option to supply an SSL certificate for RethinkDB, which will be necessary to connect to Compose unless you're fine using their SSH tunnel.

The original post mentioned SSL, but I forgot when/why it was taken out. I'll open a new issue for adding an SSL option for the RethinkDB connection as well.

danielmewes:
@levimoore Opened a new issue here: https://github.com/rethinkdb/horizon/issues/680

levimoore:
@danielmewes Thanks. I'll try to supply more information.

danielmewes:
@levimoore Thanks. Note that the Compose certificate has nothing to do with the existing SSL configuration in Horizon, or with `hz create-cert`. The Compose certificate is for connecting to RethinkDB. The existing certificate options in Horizon are for serving Horizon itself over HTTPS.

GeoffreyPlitt:
@danielmewes Now that 2.0 is out, are there any updated docs for SSL and authKey?

danielmewes:
@GeoffreyPlitt The `rdb_user` and `rdb_password` options are mentioned in the new version of the docs: https://github.com/rethinkdb/horizon-docs/blob/horizon-2.0/config-file.md . This isn't on the website yet, because Horizon 2.0 is still just a beta.
The SSL option didn't make it into 2.0.0. I think we should be able to add it in a small update, e.g. 2.1.0 or so.

GeoffreyPlitt:
Sweet, let me know if there's an issue ID I can watch

danielmewes:
@GeoffreyPlitt Just the one you've already seen, https://github.com/rethinkdb/horizon/issues/680

PR #515, opened 2016-05-26
Allow unsubscribing from all requests

### LINKED ISSUE #470: Allow unsubscribing from all requests in the client
At the moment, it appears we can only unsubscribe from `subscribe` requests in the client.  While this makes sense at first glance, `end_subscription` is implemented more generically in the server, and can be used to close any request.

I propose we allow `unsubscribe` on any request observable, which will send an `end_subscription` for that request, then ignore any subsequent responses for the given `request_id`.  This is particularly useful when users want to extend Horizon with custom request types which may involve cursors or changefeeds.

Also, we should probably rename `end_subscription` in the protocol to `close_request` or `stop_request` or something.

In the future, the server could be more proactive about cancelling requests, such as interrupting reads or writes in the database using `STOP` in the ReQL protocol, although the impacts of that should be carefully considered.

deontologician:
Fixes #470

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/515)

### DIFF
--- a/client/src/socket.js
+++ b/client/src/socket.js
@@ -219,10 +219,7 @@ class HorizonSocket extends Subject {
       // Add the request id to the request and the unsubscribe request
       // if there is one
       rawRequest.request_id = request_id
-      let unsubscribeRequest
-      if (rawRequest.type === 'subscribe') {
-        unsubscribeRequest = { request_id, type: 'end_subscription' }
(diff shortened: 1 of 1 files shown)


Tryneus:
Looks good, merging!

PR #519, opened 2016-05-26
404 response when not using --serve-static

### LINKED ISSUE #511: When not using `--serve-static`, a handler needs to be setup to correctly 404 requests
Currently if you don't have `--serve-static` set, the browser will just hang until it times out when you go to any path on the domain (except for `/horizon/*` of course) `http[s]://localhost:8181`  It should immediately 404.

Tryneus:
Fixes #511

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/519)

### DIFF
--- a/cli/src/serve.js
+++ b/cli/src/serve.js
@@ -229,6 +229,11 @@ const initialize_servers = (ctor, opts) => {
           console.info(`Static files being served from ${opts.serve_static}`);
         }
         srv.on('request', fileServer(opts.serve_static));
+      } else {
+        srv.on('request', (req, res) => {
+          res.writeHead(404);
+          res.end('404 Not Found');
+        });
       }
       srv.on('listening', () => {
         const protocol = opts.secure ? 'https' : 'http';


cli/src/serve.js
```diff
@@ -229,6 +229,11 @@ const initialize_servers = (ctor, opts) => {
           console.info(`Static files being served from ${opts.serve_static}`);
         }
         srv.on('request', fileServer(opts.serve_static));
+      } else {
```
deontologician, line 232: I could be wrong, but I think this makes it so that the server won't serve the horizon.js file

Tryneus, line 232: The server removes all existing request listeners and puts itself in their place, then passes down anything that it doesn't handle:

``` js
      const extant_listeners = server.listeners('request').slice(0);
      server.removeAllListeners('request');
      server.on('request', (req, res) => {
        const req_path = url.parse(req.url).pathname;
        if (req_path.indexOf(`${this._path}/`) === 0) {
          const sub_path = req_path.replace(path_replace, '');
          const handler = this._http_handlers.get(sub_path);
          if (handler !== undefined) {
            logger.debug(`Handling HTTP request to horizon subpath: ${sub_path}`);
            return handler(req, res);
          }
        }
        if (extant_listeners.length === 0) {
          res.statusCode = 404;
          res.write('File not found.');
          res.end();
        } else {
          extant_listeners.forEach((l) => l.call(server, req, res));
        }
      });
```

Tryneus, line 232: Although it looks like the server should be 404ing the requests already, so that's a little strange.

Tryneus, line 232: Ah, that code path no longer works since we're using `engine.io` - it installs its own request handler to pass things down (and it doesn't 404 if there's nothing there).  Since we're thinking of moving away from `engine.io`, I don't see the problem with leaving the handling in both the `cli` and `server`.

deontologician, line 232: Ok, so this does resolve the problem then.

PR #524, opened 2016-05-26
Make it more obvious that it works

oskarrough:
As I was trying out horizon for the first time my server returned a blank screen. Or so I thought because the default `<marquee>` animates in so slowly from the right. Of course, had I waited two seconds longer on my 27" screen I would have seen the 'It works!' text…

So, this proposed change reverses the direction in order for the text to appear immediately to the left.

I realize this is a minor issue, if it all, but it took me a minute to figure out.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/524)

### DIFF
--- a/cli/src/init.js
+++ b/cli/src/init.js
@@ -25,7 +25,7 @@ const makeIndexHTML = (projectName) => `\
     </script>
   </head>
   <body>
-   <marquee><h1></h1></marquee>
+   <marquee direction="left"><h1></h1></marquee>
   </body>
 </html>
 `;


dalanmiller:
@oskarrough Does it appear faster if it comes from the left versus the right? Maybe we should also increase the `font-size` by a healthy margin.

oskarrough:
Not exactly faster but if we assume most users will look left before right, it makes sense to have it start there. Increasing font-size could also solve it 👍

deontologician:
Well if we want to increase the font size, clearly we need to use the `<font>` tag

deontologician:
:lgtm: Thanks @oskarrough

PR #531, opened 2016-05-27
Maybe Typescript

deontologician:
This is an experimental thing. If it gets too nasty, this may not be merged (but we could still go to .d.ts files to declare the public api)

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/531)


joshwiens:
@deontologician - If you want a hand with the tooling to make this less "nasty" give a yell. I went the immersion route and moved my team to TypeScript for "all the things" 4 months ago. I've probably been through most of the trials you are about to run into.

deontologician:
I'm going to pause on this until Typescript 2.0 is out and we can depend on things like null checking etc.

ghost:
@deontologician Typescript 2.1 RC is out now. Any plans of looking into it again? Thanks.

Issue #492, opened 2016-05-23
Send auth token through an HTTP header rather than a query parameter


Tryneus:
See https://discuss.horizon.io/t/can-we-should-we-keep-the-token-out-of-the-url/182

The link provided is fairly compelling, and I agree that we should protect against that sort of attack.  It should be possible to send an intermediary response in the auth flow that will send the token over some custom HTTP header, like `X-Horizon-Token: <JWT>` which will store it in local storage, then redirect to the configured `auth_redirect` page.

Most of this should just be in the server's auth code, but we should also remove the code in the client that pulls the token out of the query parameters.

PR #544, opened 2016-06-01
Allow dynamic **redirect_uri**

Youmoo:
This pr makes [redirect urls](https://developer.github.com/v3/oauth/#redirect-urls) possible. Assume our auth endpoint is `/horizon/github`, when a user clicks:
- `/horizon/github?redirect_uri=/abc`, browser will redirect to `/abc` on login success
- `/horizon/github`, browser will redirect to default success url ( the entry `auth_redirect` defined in _.hz/config.toml_

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/544)

### DIFF
--- a/server/src/auth/utils.js
+++ b/server/src/auth/utils.js
@@ -107,17 +107,20 @@ const oauth2 = (raw_options) => {
   const self_url = (host, path) =>
     url.format({ protocol: 'https', host: host, pathname: path });
 
-  const make_success_url = (horizon_token) =>
-    url.format(extend_url_query(horizon._auth._success_redirect, { horizon_token }));
+  const make_success_url = (horizon_token, custom_redirect_uri) =>
+    url.format(extend_url_query(custom_redirect_uri ? 
+        url.parse(custom_redirect_uri) : 
+        horizon._auth._success_redirect, { horizon_token }));
 
   const make_failure_url = (horizon_error) =>
     url.format(extend_url_query(horizon._auth._failure_redirect, { horizon_error }));
 
   horizon.add_http_handler(provider, (req, res) => {
     const request_url = url.parse(req.url, true);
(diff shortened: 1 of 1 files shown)


Youmoo:
I have a requirement that _A user should be redirected to the previous page where he clicks the login button._ Since the login button is present on all pages,  the `auth_redirect` entry does not help.

deontologician:
ping @Tryneus

deontologician:
Hey @Youmoo, just wanted to say we'll be looking at this soon. Thanks for your patience and for submitting the PR!

Tryneus:
As it stands, this is a massive security vulnerability.  This _could_ be acceptable if we restrict the redirect URI to only the current host, but even that could be vulnerable for certain applications.

The scenario is a CSRF, and unfolds like this:
- Application is hosted on a webserver: `https://app/`
- Victim has an account with Application, authorized using Oauth provider X
- Attacker owns a webserver: `http://fake/`
- Attacker gets a victim to follow a link to the application and log in: `https://app/horizon/x?redirect_uri=http://fake/`
- Victim is automatically logged in because Application is already approved with the provider X
- Victim is redirected to Attacker's webserver with the token as a query parameter
- Attacker now has Victim's token and can now issue commands as that user until the token expires or is invalidated

We've discussed reworking how tokens are passed back to the browser in a different way - and it might be done such that we can allow something like this (see #492).  As it stands, we can't use this code though.  Thanks for the suggestion, though, @Youmoo!

sdwlig:
A CORS-like system can be used to provide secure filtering of dynamic redirects.  This is highly needed for development when the client and server sides are often developed in different environments by different people.  It is also useful so that the front-end web apps and database/server API can be in different places.

PR #556, opened 2016-06-04
Allow specifying the auth token directly to the Horizon constructor

### LINKED ISSUE #429: Modify client to allow specifying the auth token directly
Right now the only way to use an auth token you have generated is to have it in your localStorage. There is a workaround at the moment for putting on in there (`horizon.utensils.tokenStorage.set(TOKEN)`) but this is pretty hacky. It should be possible to provide a token to the Horizon client constructor, so that it ignores localStorage and uses it directly. I think we should allow the `authType` argument to receive an object

``` js
const hz = Horizon({
  authType: {
     token: "<paste jwt here>",
     storeLocally: false,
  }
})
```

If `storeLocally` is set to false, we don't even construct a TokenStorage object. This would be good for things like react native, where the only storage is AsyncStorage. It would also make it easier to use the tokens generated by `hz make-token`.

deontologician:
Resolves #429

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/556)

### DIFF
--- a/client/src/auth.js
+++ b/client/src/auth.js
@@ -33,32 +33,42 @@ export class FakeStorage {
   removeItem(a) { return this._storage.delete(a) }
 }
 
-function getStorage() {
+function getStorage(storeLocally = false) {
+  let storage
   try {
-    if (typeof window !== 'object' || window.localStorage === undefined) {
-      return new FakeStorage()
+    if (!storeLocally ||
+        typeof window !== 'object' ||
+        window.localStorage === undefined) {
+      storage = new FakeStorage()
+    } else {
+      // Mobile safari in private browsing has a localStorage, but it
+      // has a size limit of 0
+      window.localStorage.setItem('$$fake', 1)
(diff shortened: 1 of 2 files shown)


client/src/auth.js
```diff
@@ -33,32 +33,42 @@ export class FakeStorage {
...
+      storage = window.sessionStorage
     }
   }
+  return storage
 }
 
 export class TokenStorage {
   constructor({ authType = 'token',
-                storage = getStorage(),
+                storage = getStorage(authType.storeLocally),
```
hswolff, line 62: Heya! Not sure how this works? Isn't `authType` a string? How can there be a nested string? I ran into this trying out latest horizon client and noticing that client-side saving of tokens in `localStorage` no longer seems to work.

deontologician, line 62: The intention is for authType.storeLocally to be falsey (undefined) if authType is not an object. I think I just messed up the logic for this one and it should be reversed.

PR #565, opened 2016-06-06
Fix JSON fetch issue (#487).

### LINKED ISSUE #487: horizon.authEndpoint('github').subscribe returns "Unexpected token <"
```
horizon.authEndpoint('github').subscribe(endpoint => {})
e {isUnsubscribed: false, syncErrorValue: null, syncErrorThrown: false, syncErrorThrowable: false, isStopped: false…}
subscribeToResult.js:41 Uncaught SyntaxError: Unexpected token <
```

Facebook returns the same error.

Client served from server v1.0.2

poznyakovskiy:
Fixes #487

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/565)


deontologician:
The JSON fetch fix looks good, but I think we shouldn't use Allow \* by default for the server. Ideally, we could provide an argument to the server that says which endpoint static files will be served from, so we can do:

`Access-Control-Allow-Origin: http://example`

poznyakovskiy:
Updated. How about this?

deontologician:
This looks great, (I was hesitant to suggest this since it's a pain in the ass adding new config options). Pinging @Tryneus for review

server/src/server.js
```diff
@@ -140,6 +140,7 @@ class Server {
 
     this.add_http_handler('auth_methods', (req, res) => {
       res.writeHead(200, { 'Content-Type': 'application/json' });
+      res.writeHead(200, { 'Content-Type': 'application/json', 'Access-Control-Allow-Origin': opts.access_control_allow_origin });
```
dalanmiller, line 142: Why set `Content-Type` twice? Did you forget to take out the previous `writeHead`?

dalanmiller:
If we are allowing another configuration option I think we should enforce also putting it into the checks for environment variables and also add a command-line flag. 

It could get a bit troublesome to check whether or not a configuration option is only implemented in one or only some of the configuration sources. i.e. only in config files but no check for the same option in the command-line parser or environment variables check. Not to mention confusing for someone expecting the flag to be there after seeing the option exists in the config file. 

Let me know if you want to tackle this @poznyakovskiy or else I can add the rest of the checks in `cli/src/serve.js`.

poznyakovskiy:
Not sure where checks would go to. Can you give me a hint?

marshall007:
@dalanmiller IMO, this doesn't really seem like one of those options worth exposing a command-line flag for. It just adds bloat to the CLI's available arguments. Instead, I think using one of the higher-level flags like `--dev` should automatically set `Access-Control-Allow-Origin: *` unless already set in the `config.toml`.

deontologician:
Going to try to merge this soon since it's a bad bug. ping @Tryneus

PR #586, opened 2016-06-14
Fix lazy currentUser

### LINKED ISSUE #567: hz.currentUser doesn't execute until something else is executed
- Server version v1.1.1
- Client version The same I think

``` js
const hz = Horizon();
hz.currentUser().fetch().subscribe( (user) => console.log(JSON.stringify(user)) );
```

^ this example does not work unless you perform another query or explicitly do `hz.connect()`

deontologician:
Fixes #567 

The issue was that subscribing to the handshake doesn't force a connection, it only sets a callback whenever the handshake happens to be completed. Now, when subscribing `currentUser().fetch()` or `currentUser.watch()` an actual connection will be made.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/586)

### DIFF
--- a/client/src/ast.js
+++ b/client/src/ast.js
@@ -6,6 +6,7 @@ import 'rxjs/add/operator/scan'
 import 'rxjs/add/operator/filter'
 import 'rxjs/add/operator/map'
 import 'rxjs/add/operator/toArray'
+import 'rxjs/add/operator/take'
 
 import snakeCase from 'snake-case'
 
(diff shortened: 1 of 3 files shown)


client/src/ast.js
```diff
@@ -355,3 +356,39 @@ export class Limit extends TermBase {
     super(sendRequest, query, [])
   }
 }
+
+
+export class UserDataTerm {
```
deontologician, line 361: This is just moved from index.js where it didn't really belong

Issue #373, opened 2016-05-17
Plugins


bopjesvla:
Reading through the discussion on document validation, it seems to me that if everything is implemented as a fixed API, horizon will only be used by people with specific use cases.

I don't think document validation should be part of the core, for the simple reason that no one solution will suit every app. You might be thinking about storing the schemas in the database. What if I want to validate login information before the web socket is opened? The schemas might be read from JSON files. What if I want to use YAML? You might pick ajv as the json-schema implementation because it's fast. What if I want to use jsen because it's three times as small? What if my app doesn't need any fancy APIs whatsoever, and I'm just using horizon for the realtime feeds?

---

This is what I imagine a document validation plugin would look like:

``` js
let validate = require("json-something").validate
module.exports = function(hz, opts) {
    let schemas = hz("schemas").fetch();
    let validate = (table, doc) => validate(schemas[table], doc)
    hz.on("insert", validate)
    hz.on("update", validate)
}
```

An auth plugin:

``` js
module.exports = function(hz, opts) {
    let users = hz("users");
    if (!users) {
        // initialize
    }
    hz.on("connect", credentials => users.findOne({user: credentials.user, password: credentials.password}))
}
```

Plugin configuration would be done in config.toml and/or a configuration module like webpack's config.js.

bopjesvla:
> I think if you're interested in the different between size and speed of json-schema parsing libraries, then you're in the realm of writing your own custom backend.

That particular comment was about document validation on the client. And yeah, it wasn't the best example. The first and the last one are more important.

A slightly more concrete proposal that depends on #345:
- Every X in `[plugins.X]` in `config.toml` is required and called with the horizon instance and config options. This can also be done programmatically from `server.js`.
- The horizon server exposes events which plugins and code in `server.js` can hook into.
- The client also exposes things like insert hooks, which client-side code can hook into.

I'll expand on why I think auth and validation should be implemented as plugins later, but I'd like to hear your thoughts on this first.

bopjesvla:
Disorganised notes:
- Unofficial plugins have the luxury of making assumptions about what stack they're used in.
- Lacking that luxury, horizon's APIs are forced to make assumptions about the nature of the app they're used in.
- The most simple use case for horizon is public real-time display of data, which requires authentication nor document validation.
[...]

Issue #368, opened 2016-05-16
Worker processes


deontologician:
Horizon has no concept of worker processes currently. We should add some way of doing worker processes.

Probably should use changefeeds on some kind of job table to ensure jobs are only worked by one server at a time.

deontologician:
I think this is subsumed by #311. Closing

Issue #311, opened 2016-04-30
Ideation: Scheduled tasks


jeveloper:
Hello everyone,

Lightweight jobs

There are plenty of scenarios where you'd like to run jobs, sure this can be accomplished via queues and such, however given that meteor (https://github.com/vsivsi/meteor-job-collection/ ) uses a mongo collection as job store (using prefixes , priorities, delays, repeats, failed attempts, etc) i was wondering if horizon with rethinkdb should have something like this.

This should bring us to the topic of workers, would there be a concept like that on horizon server side or not. I'd say a lightweight way is to use node's workers+cluster , its fairly simple , however scales only by amount of virtual cores.

thank you RethinkDb team

deontologician:
Related to #368

Issue #216, opened 2016-03-30
Get email/name/etc from OAuth providers


deontologician:
My understanding is that right now we get no information in Horizon from the providers, just true/false "person has an account here"

ping @Tryneus to verify that we don't already get this info, and thoughts on how to implement it

steelbrain:
Bump

0xgeert:
> The provider given data should be read-only, and we should be able to specify which data to get from the provider in the toml config

This would presumably by done by asking for oauth scopes?

Related: would be great to be able to (incrementally) ask for additional scopes only when needed. E.g to only ask for full blown read/write access when an authenticated user actually intends to use a feature that requires said scope.

daniel1943:
Any progress on this?

dalanmiller:
Hey @daniel1943, Plugins is scheduled for ~~either 2.0 or~~ 2.1 and the expanded OAuth functionality will come in shortly thereafter. We'll keep you posted 👍 and if you'd like to hear right away, definitely follow the Announcements category over at https://discuss.horizon.io.

natew:
This one is pretty huge for us. Horizon isn't really useable for us without being at least able to tie email back to account, and it seems it also doesn't have support for wiring a 3rd party account system.

deontologician:
I bumped this up to post 2.0 release

sp90:
Bump, whats the status on this, anyone working on it?

PR #588, opened 2016-06-14
Plugins RFC

deontologician:
[Rendered version of RFC](https://github.com/rethinkdb/horizon/blob/rfc/plugins/rfcs/plugins.md)

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/588)

### DIFF
--- a/rfcs/plugins.md
+++ b/rfcs/plugins.md
@@ -126,7 +126,7 @@ will look like:
 }
 ```
 
-### Http route
+### HTTP route
 
 The `httpRoute` plugin property is an object with a single key
 defining the http route the plugin will receive requests from. The
@@ -174,9 +174,6 @@ to process titles, and values which are objects with 2 properties:
  - `cleanup`: a function receiving the plugin state that cleans up the
    process
 
-**TODO**: do we need to provide process monitor options? Stuff like
-backoff values and restart limits etc.
-
 ### Environment vars
 
 The `env` plugin property is an object from keys with environment
@@ -202,7 +199,7 @@ variables like `$PATH` etc, and doesn't want to mutual exclusivity to be enforce
 
 The `validationHelpers` plugin property is an object containing
 functions that will be made available in the validation context. There
-are no restrictions on the types of the functions.
+are no restrictions on the types of the functions here
 
 Example:
 ```js
@@ -211,15 +208,102 @@ validationHelpers: {
 }
 ```
 
+This will make the `isTruthy` function available to the `validators`
+functions specified in `.hz/config.toml`:
+
+```toml
+[groups.authenticated.rules.foo]
+validator = """
+  function(context, oldValue, newValue) {
+    return isTruthy(newValue.isFoo)
+  }
+"""
+```
+
 ### Requests
+
+The `requests` plugin property allows specifying new request types
+that can be sent by the client. The keys are the name of the new
+requests, the value is an object with the following properties:
+  - `optionsSchema`: A Joi schema to validate the request type.
+  - `handler`: A function that receives a validated request, the
+    current plugin state
+
+The schema provided just validates the contents of the `options`
+field. So for a schema like:
+
+```js
+requests: {
+  foo: {
+    optionsSchema: {
+      bar: Joi.boolean().default(false),
+    },
+    ...
+  }
+}
+```
+
+A matching raw request might look like:
+
+```js
+{
+    "request_id": 33,
+    "type": "foo",
+    "options": {
+        "bar": true
+    }
+}
+```
 
 ### Activate
 
+The `activate` plugin property must be present and contain a function
+that accepts an object with the following properties:
+  - `config`: config options from the config.toml section for the
+    plugin. This includes user overrides which are not specified in
+    the `config` plugin property.
+  - `env`: an object with relevant environment variables (defined in
+    [environment](#environment))
+  - `rdbConn`: a function that returns an open rethinkdb connection
+    - If the plugin doesn't need a rethinkDB connection, it shouldn't
+      call this function
+
+The function should return an object which will be passed to various
+other callbacks for the plugin and contains all internal state the
+plugin needs to function. This includes keeping a reference to the
+rethinkDB connection if necessary.
+
 ### Deactivate
 
-The plugin is required to completely clean itself up in the
-`deactivate` function, leaving no state behind.
+The `deactivate` plugin property must be present and contain a function that accepts the current plugin state and frees any resources the plugin is using. (Disconnect any connections, close any files, etc)
 
 ## User control over plugins
 
-- Users may override the httpRoute for any plugin
+Users may override some names provided by the plugin:
+  - rename the [httpRoute](#http-route) if defined
+  - rename any function in [validationHelpers](#validation-helpers)
+  - rename any new [request types](#request-types)
+
+Users configure the plugins with a section like:
+
+```toml
+[plugin.plugin-name]
+httpRoute = "httpRouteOverride",
+validationHelpers = {
+ isTruthy = "namespacedIsTruthy"
+}
+requests = {
+  foo = "namespacedFoo"
+}
+# plugin specified options ...
+```
+
+Every plugin can always accept the `httpRoute`, `validationHelpers`,
+and `requests` keys, even if the plugin specifies no [config](#config)
+section. Horizon will respect these renames itself, the plugin doesn't need to do anything based on them. It will still be passed the options though in the [activate](#activate) function, so it can optionally do something if it wants to.
+
+# Open questions:
+
+- Can you instantiate a plugin more than once per app?
+- Do we need to provide process monitor options? Stuff like backoff
+values and restart limits etc.
--- a/rfcs/plugins.md
+++ b/rfcs/plugins.md
@@ -6,7 +6,7 @@ of configuration.
 
 Plugins are intended to be flexible enough to implement things like:
 
- - GraphQL endpoints (#125)
+ - GraphQL endpoints #125)
  - REST endpoints (#510)
  - Custom horizon commands (#337)
  - Custom authentication methods (#428)
@@ -24,18 +24,128 @@ Plugins are intended to be flexible enough to implement things like:
  - NPM modules with the keyword 'horizon-plugin'
  - export an object with the name 'plugin' that has one or more of the
    following properties defined on it:
+   - `name`: a name for the plugin. Defaults to the package name
    - `commands` for extending the `hz` command line tool [commands](#commands)
    - `processes`: specifies subprocesses to spawn [processes](#processes)
    - `endpoints`: specifies http endpoints that will be owned by the plugin [endpoints](#endpoints)
    - `config`: specifies new configuration options [config](#config)
+   - `env`: specifies environment variables the plugin cares about [environment vars](#environment-vars)
    - `requests`: specified new request types to accept over the websocket connection [requests](#requests)
-   - `activate`: Called once before the plugin is activated
-   - `deactivate`: Called once before the plugin is deactivated
- - The plugin is required to completely clean itself up in the
-   `deactivate` function, leaving no state behind.
+   - `activate`: Called once before the plugin is activated [activate](#activate)
+   - `deactivate`: Called once before the plugin is deactivated [deactivate](#deactivate)
 
 ### Commands
-### Processes
-### Endpoints
+
+The `commands` plugin property must be an object with the following
+properties:
+  - `name`: a string with the name of the command to
+    implement. Defaults to an underscore-to-hyphen converted version
+    of the npm package name of the plugin
+  - `helpText`: a short description of what the command does. Mandatory.
+  - `addArguments`: a function that receives the
+    [argparse parser](http://nodeca.github.io/argparse/#HelpFormatter.prototype.addArgument)
+    for the current command, and adds any options to that parser. This
+    can include creating
+    [subparsers](http://nodeca.github.io/argparse/#ArgumentParser.prototype.addSubparsers)
+    - default is a no-op function
+  - `processConfig`: a function that receives:
+   1. the final result of the parsed options from the command line
+   2. an object representing the plugin's [config options](#config) that were in the config.toml file
+   3. an object with the values of any environment variables the plugin cares about
+    any. It returns the final options, merging command line and config
+    options any way it wants to.
+    - default is a function that merges the config, env and command
+      line arguments in a precedence order the same way built in
+      commands do.
+  - `runCommand`: a function that receives the merged result of
+    `processConfig` as its first argument and a `done` function as its
+    second argument. It should execute its task, calling `done` with an error if something goes wrong.
+    - Note: it should *not* call `process.exit`.
+    - Mandatory option, since this is the entire point of the command.
+
+Note that the command will not ever have `activate` or `deactivate`
+called, those functions are for server plugins.
+
 ### Config
+
+The `config` plugin property specifies which config options
+will be forwarded from the config file section for the plugin.
+
+So a `config` key like:
+
+```js
+{
+  name: 'my-plugin-name',
+  config: {
+    someKey: 'boolean',
+    some_other_key: 'string',
+  }
+}
+```
+
+Will allow the following section in `config.toml`:
+
+```toml
+[plugin.my-plugin-name]
+someKey = true
+some_other_key = 'some kinda string'
+```
+
+The values from this config section will be passed to both the
+`processConfig` function for commands, as well as the
+[`activate`](#activate) function for server plugins. The object passed
+will look like:
+
+```js
+{ someKey: true, some_other_key: 'some_kinda_string' }
+```
+
+### Endpoints
+
+The `endpoints` plugin property is an object with keys that correspond
+to endpoints the plugin will be responsible for, and values that are
+functions that handle requests to that endpoint. The functions receive
+three arguments:
+ - An Express `request` object
+ - An Express `response` object
+ - A state object of whatever kind is returned by the plugin's
+   [activate](#activate) function
+
+The handler should deal with whatever HTTP methods come to the route.
+
+Example:
+
+```js
+{
+  endpoints: {
+    'hello-world': (req, res, state) => {
+      res.send(`Hello ${state.value}`);
+      res.end()
+    },
+    'set-world': (req, res, state) => {
+      state.value = req.body
+      res.end()
+    }
+  }
+}
+```
+
+Notes:
+ - The endpoint is responsible for calling `res.end()`.
+ - The route is a hard-prefix from the root of the horizon server address.
+  - It cannot clash with any other plugin routes (no overlapping)
+  - It cannot clash with the route horizon itself (usually `"horizon"`)
+ - Express 4.0
+
+### Processes
+
+### Environment vars
+
 ### Requests
+
+### Activate
+
+### Deactivate
+
+The plugin is required to completely clean itself up in the
+`deactivate` function, leaving no state behind.
--- a/rfcs/plugins.md
+++ b/rfcs/plugins.md
@@ -17,7 +17,7 @@ Plugins are intended to be flexible enough to implement things like:
 ## Things plugins can do
  - Add new configuration options
  - Add new `hz` commands
- - Add new http endpoints
+ - Add a new http route
  - Add new helper functions available to validators
  - Spawn supervised processes
  - Add new horizon request types useable by the client
@@ -28,7 +28,7 @@ If you need to do any of these things, writing a custom backend and
 importing `@horizon/server` as a module will be your best bet.
 
  - hook into websocket request types it didn't define
- - respond to requests to http endpoints it didn't define
+ - respond to requests to http routes it didn't define
  - modify state of other plugins
  - override validation results
  - see configuration options for other plugins
@@ -48,7 +48,7 @@ specification (say if two plugins take over the same http path).
    - `name`: a name for the plugin. Defaults to the package name
    - `command` for extending the `hz` command line tool [command](#command)
    - `processes`: specifies subprocesses to spawn [processes](#processes)
-   - `endpoints`: specifies http endpoints that will be owned by the plugin [endpoints](#endpoints)
+   - `httpRoute`: specifies the http route that will be owned by the plugin [Http route](#http-route)
    - `config`: specifies new configuration options [config](#config)
    - `env`: specifies environment variables the plugin cares about [environment vars](#environment-vars)
    - `requests`: specified new request types to accept over the websocket connection [requests](#requests)
@@ -126,11 +126,11 @@ will look like:
 }
 ```
 
-### Endpoints
+### Http route
 
-The `endpoints` plugin property is an object with keys that correspond
-to endpoints the plugin will be responsible for, and values that are
-functions that handle requests to that endpoint. The functions receive
+The `httpRoute` plugin property is an object with a single key
+defining the http route the plugin will receive requests from. The
+value is the request handler for the route. The function receives
 three arguments:
  - An Express `request` object
  - An Express `response` object
@@ -142,32 +142,28 @@ The handler should deal with whatever HTTP methods come to the route.
 Example:
 
 ```js
-{
-  endpoints: {
+httpRoute: {
     'hello-world': (req, res, state) => {
-      res.send(`Hello ${state.value}`);
-      res.end()
+        res.send(`Hello ${state.value}`);
+            res.end()
     },
-    'set-world': (req, res, state) => {
-      state.value = req.body
-      res.end()
-    }
-  }
 }
 ```
 
 Notes:
- - The endpoint is responsible for calling `res.end()`.
+ - The handler is responsible for calling `res.end()`.
  - The route is a hard-prefix from the root of the horizon server address.
-  - It cannot clash with any other plugin routes (no overlapping)
-      - If multiple routes need to be handled, they have to all start
-        with the same prefix
+  - It cannot clash with any other other loaded plugin routes (no
+    overlapping)
+      - If multiple routes need to be handled, the r
       - Example, if you want to create a REST plugin, you need to
-        create an endpoint like `rest`, and handle the suffixes in
-        your function, rather than registering many different
-        endpoints like `/users`, `/blogs` etc.
-  - It cannot clash with the route horizon itself (usually `"horizon"`)
- - Express 4.0
+        specify the route as something like `rest` and the endpoints
+        underneath it like `/rest/users` `/rest/vehicles` etc.
+  - It cannot clash with the route horizon uses itself (usually
+    `"horizon"`)
+ - In the configuration, the user may override the httpRoute for the
+   plugin, and the plugin must behave properly.
+ - Express 4.0 api for request and response objects
 
 ### Processes
 
@@ -223,3 +219,7 @@ validationHelpers: {
 
 The plugin is required to completely clean itself up in the
 `deactivate` function, leaving no state behind.
+
+## User control over plugins
+
+- Users may override the httpRoute for any plugin
--- a/rfcs/plugins.md
+++ b/rfcs/plugins.md
@@ -246,7 +246,8 @@ field. So for a schema like:
 requests: {
   myRequestType: {
     optionsSchema: {
-
+      collection: Joi.string(),
+      findAll: Joi.array(Joi.any()),
       myRequestType: Joi.array().ordered(
           Joi.number(), Joi.boolean().default(false))
     },
--- a/rfcs/plugins.md
+++ b/rfcs/plugins.md
@@ -222,9 +222,20 @@ validator = """
 ### Requests
 
 The `requests` plugin property allows specifying new request types
-that can be sent by the client. The keys are the name of the new
-requests, the value is an object with the following properties:
+that can be sent by the client. The keys of the provided object are
+the name of the new requests, the values are an object with the
+following properties:
   - `optionsSchema`: A Joi schema to validate the request type.
+  - `clientValidation`: An object with the keys:
+    - `minArgs`: minimum number of arguments accepted
+    - `maxArgs`: maximum number of arguments accepted
+    - `legalToChainFrom`: a list of terms that can be chained before
+      this request method. The special value `"/"` indicates the
+      method is available directly on the `horizon`
+      instance. `"collection"` indicates it's chainable from a
+      collection.
+    - `nullable`: a boolean determining whether any arguments can be
+      null
   - `handler`: A function that receives a validated request, the
     current plugin state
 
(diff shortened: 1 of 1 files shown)


rfcs/plugins.md
```diff
@@ -0,0 +1,151 @@
...
+   - `endpoints`: specifies http endpoints that will be owned by the plugin [endpoints](#endpoints)
+   - `config`: specifies new configuration options [config](#config)
+   - `env`: specifies environment variables the plugin cares about [environment vars](#environment-vars)
+   - `requests`: specified new request types to accept over the websocket connection [requests](#requests)
+   - `activate`: Called once before the plugin is activated [activate](#activate)
+   - `deactivate`: Called once before the plugin is deactivated [deactivate](#deactivate)
+
+### Commands
+
+The `commands` plugin property must be an object with the following
```
Tryneus: It sounds like `commands` should be an object or array of objects.

deontologician: I thought of that, but then I thought if they wanted to implement multiple commands, they could just do them as subcommands. So instead of `hz crom-input` and `hz crom-output` they'd do `hz crom input` and `hz crom output`

rfcs/plugins.md
```diff
@@ -0,0 +1,151 @@
...
+  - `addArguments`: a function that receives the
+    [argparse parser](http://nodeca.github.io/argparse/#HelpFormatter.prototype.addArgument)
+    for the current command, and adds any options to that parser. This
+    can include creating
+    [subparsers](http://nodeca.github.io/argparse/#ArgumentParser.prototype.addSubparsers)
+    - default is a no-op function
+  - `processConfig`: a function that receives:
+   1. the final result of the parsed options from the command line
+   2. an object representing the plugin's [config options](#config) that were in the config.toml file
+   3. an object with the values of any environment variables the plugin cares about
```
Tryneus: Shouldn't the plugin just check the environment variables directly, or are we going to try to sandbox the plugins?

deontologician: I mainly put this in here so that we can determine at startup if any plugins are looking at the same environment variables and disallow it.

mlucy:
One piece of feedback: I think we should require people to specify a namespace for their plugin, and then namespace things like environment variables and endpoints and whatnot.  (So if I was making a GraphQL plugin, I might register `gql` as my namespace, and then all my endpoints would be `/horizon/gql/foo` and all my environment variables would be `GQL_BAR` and so on.)  Otherwise plugins are inevitably going to conflict with each other, and it's going to be very annoying for users to deal with.

rfcs/plugins.md
```diff
@@ -0,0 +1,151 @@
...
+
+```js
+{
+  name: 'my-plugin-name',
+  config: {
+    someKey: 'boolean',
+    some_other_key: 'string',
+  }
+}
+```
```
Tryneus, line 187: Maybe we should just use `Joi` for this, since it's how we're doing schema validation elsewhere.  We won't have to reimplement a bunch of stuff.

deontologician:
@mlucy While I didn't mention an explicit namespacing scheme, the plugin system should verify at plugin activation time whether anything is in conflict. This would allow people to use nicer names for things. Essentially the idea is that the plugins are somewhat less paranoid about name conflicts since they declare everything they use. So there might be multiple plugins that provide a `/graphQL` endpoint, but you wouldn't ever have both of them installed at the same time.

rfcs/plugins.md
deontologician, line 187: That's a good idea, I'll change it

rfcs/plugins.md
```diff
@@ -0,0 +1,151 @@
...
+    implement. Defaults to an underscore-to-hyphen converted version
+    of the npm package name of the plugin
+  - `helpText`: a short description of what the command does. Mandatory.
+  - `addArguments`: a function that receives the
+    [argparse parser](http://nodeca.github.io/argparse/#HelpFormatter.prototype.addArgument)
+    for the current command, and adds any options to that parser. This
+    can include creating
+    [subparsers](http://nodeca.github.io/argparse/#ArgumentParser.prototype.addSubparsers)
+    - default is a no-op function
+  - `processConfig`: a function that receives:
```
mlucy: I think we should omit `processConfig` (and maybe simplify `addArguments` to be more restrictive).  I think we should try to force plugins to follow certain standards for what arguments they accept, specifically that all arguments can be specified with the same name in the config file, in environment variables, or on the command line (in increasing order of precedence).  In practice every good plugin will end up doing that anyway, so why allow for bad plugins with confusing argument semantics like giving the config file precedence over the command line or using different names for the environment variable and the command line option?

mlucy:
Is the idea that Horizon will look in a certain directory on startup to figure out what plugins to load?  Is it OK for different Horizon nodes to be running different sets of plugins?  (Also, am I correct in assuming that plugins live on disk rather than inside RethinkDB?)

mlucy:
@deontologician -- I'm more worried about e.g. a GraphQL plugin and a performance graphing plugin both providing a `graphFOO` endpoint.  Or two other plugins with totally unrelated behavior that happen to pick the same name for an endpoint.

I think we should have namespacing for the same reason people who write packages for programming languages don't get to just dump their functions into the global namespace: if people put things into the global namespace, conflicts are _literally_ inevitable once there are a certain number of plugins, and it's extremely frustrating to not be able to use two totally unrelated plugins together just because they happened to pick the same name for something.  (You could argue that a responsible plugin writer would namespace their endpoints _anyway_, which is 100% true, but if that's what a responsible plugin writer would do, why not just force everyone to be responsible?)

rfcs/plugins.md
deontologician: `addArguments` is for specifying the command line syntax, whereas `processConfig` runs after the command has been run and the options are parsed out successfully. I don't have a strong opinion, but it seems like environment variables may need more namespacing than config options or commands. Command options can be part of subcommands and the config has sections which are both like namespaces.

It'd be nice if we didn't enforce a namespacing scheme on the environment variables, or ensure that they have a specific precedence order, but maybe it's not a useful kind of flexibility.

deontologician:
@mlucy in the case where two plugins pick the same name for an endpoint, the plugins should cooperate and pick different names. I think this kind of conflict will be pretty rare, and as long as we don't allow them both to run (by checking that endpoints don't overlap) it won't be a problem. It's much better to have clean looking paths and command names than to namespace everything in case two plugins happen to have a name clash.

I'm not arguing a responsible plugin writer should namespace their endpoints, I think a responsible plugin writer should name the endpoint whatever is most natural, and if/when a conflict occurs, decide if the newer project should rename its endpoints or namespace them (and they'll probably always just rename them). Or they could decide the conflict is ok since usage of the two plugins should be mutually exclusive (like two graphQL plugins)

deontologician:
> Is the idea that Horizon will look in a certain directory on startup to figure out what plugins to load?

Node modules have a field in their package.json that allows them to specify keywords. So we'll assume any modules currently installed that have the correct keyword are horizon plugins and try to load them. This is how another module project, [Yeoman](http://yeoman.io/authoring/), detects plugins and it seems reasonable to me.

> Is it OK for different Horizon nodes to be running different sets of plugins?

My gut feeling is no for simplicity's sake, but it might be worth allowing so people can add worker plugins only on certain servers, things like that.

> am I correct in assuming that plugins live on disk rather than inside RethinkDB?

yes, they'll live in the node_modules directory

marshall007:
Just at first glance, I think this RFC is trying to squeeze too much under the "plugin" umbrella. There are more or less four basic ways one might want to extend the Horizon server:
1. Extending the WebSocket protocol.
2. Hooking into life-cycle events (i.e. for auth, request validation, document validation, etc).
3. Adding custom HTTP endpoints.
4. Workers/Scheduled Tasks

I'm not sure it makes sense for a single "plugin" to be responsible for configuring all these things at once. [Primus](https://github.com/primus/primus) has done really well in this regard and is a pretty solid example for many of the goals in this RFC. Several core components in Primus are in fact implemented as plugins/middleware, which I think is along the same lines as how you guys are thinking about this. Notice how they have different mechanisms for hooking into the server (authorization, middleware, plugins, events) and each is tailored to a particular set of use-cases.

Likewise, I don't think "`command` plugins" should be handled here since that seems to imply a hard dependency on the CLI. You should be able to write plugins that can be used by any `@horizon/server` instance regardless of how you spin it up. Why not have the Horizon CLI be independently extensible?

deontologician:
> Likewise, I don't think "command plugins" should be handled here since that seems to imply a hard dependency on the CLI.

We haven't totally fleshed out how plugins should be integrated if you're embedding the server, but I think it's reasonable to just have the cli aspect be a no-op in that case. It's kind of nice if you have a server plugin to be able to add related convenience commands to the cli (like create an example config), and it's probably not worth having multiple plugin types.

If you're doing manual backend and a plugin only has a cli component, you just wouldn't use that plugin, so it's not a huge burden.

The idea of linking them together is to make it clear they're intended to be significant fully featured extensions to Horizon, rather than disconnected addons. In addition, the plugin will be able to share its own state that is isolated from other plugins, so if several of these pieces need to work together, they can. For example, if a GraphQL plugin wanted to define both an http endpoint and a new Horizon request type it could share configuration and state between the two.

marshall007:
> The idea of linking them together is to make it clear they're intended to be significant fully featured extensions to Horizon, rather than disconnected addons.

@deontologician right, this makes sense but there are two cases to consider:
1. Extending with one-off "addons".
2. Installing a full-fledged third-party plugin.

I think (1) is going to be the more common case and I wouldn't want to create a full plugin definition just to add a couple HTTP endpoints. What I'm getting at is that I would prefer a solution that provides specialized "addon" APIs for integrating into the server in various ways and then a "plugin definition" just exports some metadata and a handler function that receives a server instance to configure.

Rough example:

``` js
const plugin = function (server, options, next) {
  // add HTTP endpoint
  server.http.endpoint('/test', (req, res) => {
    res.send(...);
  });

  // extend WS protocol
  server.ws.command('test', ...);

  // more configuration here...

  next();
};

plugin.meta = {
  name: 'my-plugin',
  version: 'v0.0.1',
  config: Joi.object() // schema for `options`, validated before handler is called
};

module.exports = plugin;
```

deontologician:
>  a "plugin definition" just exports some metadata and a handler function that receives a server instance to configure.

I think fundamentally we should avoid any kind of free-form configuration like this. It's more flexible, but I think it's better to make it hard for plugins to step on each other's toes than it is to make them completely free to add any functionality to the server. Specifically, I think we should avoid the middleware paradigm with plugins, leaving middleware kinds of setups to users who are embedding horizon

rfcs/plugins.md
deontologician: I added a note to the env section explicitly mentioning you can just check process.env if you want, but if you want the plugin system to check that you "own" some env variables, you'd specify them here

bopjesvla:
> I think fundamentally we should avoid any kind of free-form configuration like this. It's more flexible, but I think it's better to make it hard for plugins to step on each other's toes than it is to make them completely free to add any functionality to the server. Specifically, I think we should avoid the middleware paradigm with plugins, leaving middleware kinds of setups to users who are embedding horizon

How about exposing a wrapper around the horizon server with limited functionality? (It'd be great if this wrapper was isomorphic to the horizon client, but I don't know if that's doable.)

Also, as I noted in #373, I really think authentication and validation should be implemented through plugins hooking into onconnect and oninsert/onupdate events, respectively:

```
// plugin

module.exports = function(hz) { // taking a wrapped hz server
    hz.on("connect", creds => {
        if (!hz.find({user: creds.user, pw: creds.pw}))
            // A special error that is magically propagated to the client, like Meteor's Meteor.Error   
            throw new hz.Error("invalid-creds", "Incorrect username or password.")
    })
}

// client

hz.connect({user: "erdogan", pw: "merkel1234"})
```

marshall007:
> Specifically, I think we should avoid the middleware paradigm with plugins, leaving middleware kinds of setups to users who are embedding horizon

@deontologician I actually completely disagree. For one thing I think Node developers are used to this paradigm and it is adopted by virtually all popular server frameworks. More importantly, though, I think the flexibility it provides generally leads to a much richer plugin ecosystem. You can imagine plugins that simply provide additional abstractions for things like configuring validation or a plugin that would fully integrate PassportJS authentication.

These kinds of things would be nearly impossible to do under a structured and isolated plugin interface.

deontologician:
> I think Node developers are used to this paradigm and it is adopted by virtually all popular server frameworks

This is true, and if you're a node developer I think it's reasonable to build your app this way. The plugin system I'm proposing isn't for node developers, it's for people using horizon in serverless apps who would like to get additional functionality. The idea is that plugins as described here are more like lego pieces than middleware because a lot of the stuff they provide is already available to node developers writing their own app. I think it would be a nice bonus to make it easy for node developers to be able to use horizon plugins to save some work, but it's a secondary or tertiary goal here.

> More importantly, though, I think the flexibility it provides generally leads to a much richer plugin ecosystem. [... Some] things would be nearly impossible to do under a structured and isolated plugin interface.

This is true, and expected for this plugin interface. If you need absolute flexibility this plugin system won't provide it. Embedding horizon is a good option if you find a plugin isn't flexible enough. But by making the interface of these plugins more limited but declarative (stating up front which commands they add, which routes they create and which request types they insert) we can make it easier to keep plugins from breaking horizon and from stepping on each other in a way that a frontend developer will have a hard time debugging.

With middleware you need to have an intimate understanding of how each piece interacts with the others, the order its effects are applied in matters, and other things that are a huge distraction from the backendless experience. Middleware doesn't automatically check that it isn't interfering with other middleware, and it's just a bunch of functions so there's no real way to automatically determine anything about how it works. In my opinion this makes it a bad model for a plugin system which should "just work".

mglukhovsky:
This raises a question that isn't clear to me from the RFC: when you move your app from `hz serve` to an embedded server in Node.js, are you able to continue using Horizon plugins?

deontologician:
It's not documented here, since this is mostly about defining the api plugins writers will need to adhere to, rather than the api of the server. It should be possible, but since (as the rfc currently is) we have a dependency on Express request/response objects. We could provide a compatibility library though to allow plugins in  node apps that embed Horizon

abastardi:
> It should be possible, but since (as the rfc currently is) we have a dependency on Express request/response objects.

Does that mean they will at least work in an Express app?

deontologician:
Not automatically, since the @horizon/server module won't include express itself. We'll have to add some kind of adapter module (possibly something that will fall out of implementing this RFC), and that will be responsible for forwarding on requests. It should be simple for that adapter to add itself as an express middleware and multiplex requests to plugin endpoints.

mlucy:
Is the dependency on Express request/response objects necessary?  It might be easier to just pass our own data structures around, so that it's painless to move to embedding the server in a node app, and offer a shim to/from Express in places where it's necessary.

rfcs/plugins.md
mlucy: Ah, OK.

Personally I would:
- Only let plugins claim environment variables prefixed with their name (or a shortname or something).  So an Elasticsearch plugin could register `ELASTICSEARCH_PORT` but not just plain old `PORT`.  I think that would be clearer to users and less prone to conflicts.  (And arguable more consistent since the config sections are acting basically like namespacing.)
- Enforce the command line beats environment beats config ordering.  I think it would be _extremely_ confusing if different plugins were using different precedence schemes.

rfcs/plugins.md
```diff
@@ -0,0 +1,349 @@
...
+    },
+}
+```
+
+Notes:
+ - The handler is responsible for calling `res.end()`.
+ - The route is a hard-prefix from the root of the horizon server address.
+  - It cannot clash with any other other loaded plugin routes (no
+    overlapping)
+      - If multiple routes need to be handled, the r
```
mlucy: I think this got truncated.

rfcs/plugins.md
```diff
@@ -0,0 +1,349 @@
...
+ - In the configuration, the user may override the httpRoute for the
+   plugin, and the plugin must behave properly.
+ - Express 4.0 api for request and response objects
+
+### Processes
+
+The `processes` plugin property is an object with keys that correspond
+to process titles, and values which are objects with 2 properties:
+ - `run`: a function receiving the plugin state that runs as the main
+   thread of the process. It should not spawn processes itself
```
mlucy: I'm a little confused by this interface.  If I want to e.g. start a RethinkDB proxy in the background, how would I do that except by spawning a process in `run`?

rfcs/plugins.md
```diff
@@ -0,0 +1,349 @@
...
+The values of env variables that successfully validate are passed to
+the [activate](#activate) function as well as to the `processConfig` function for any [command](#command).
+
+**Note**:
+
+The env vars declared here will be validated to ensure no other plugin
+declares them. If a plugin wants to make use of externally defined env
+variables like `$PATH` etc, and doesn't want to mutual exclusivity to be enforced by Horizon, it can just access `process.env` like normal.
+
+### Validation helpers
```
mlucy: Is it possible to write a validation helper that interacts with the outside world?  Like talk to another process, or access data that the plugin is reading from RethinkDB in the background?

deontologician:
> Is the dependency on Express request/response objects necessary?

Probably not, but we need to expose some interface, so the question is who to emulate. We could come up with our own interface, but that doesn't seem useful, we'd probably end up just renaming express method names. If there's a solid proposal that makes sense though, let's do it. I think we need headers, body writers, and some way of completing the response (for async stuff)

rfcs/plugins.md
deontologician: In all likelihood, I got distracted and just stopped typing

rfcs/plugins.md
```diff
@@ -0,0 +1,349 @@
...
+```
+
+### Requests
+
+The `requests` plugin property allows specifying new request types
+that can be sent by the client. The keys of the provided object are
+the name of the new requests, the values are an object with the
+following properties:
+  - `optionsSchema`: A Joi schema to validate the request type.
+  - `clientValidation`: An object with the keys:
```
mlucy: I'm not sure doing argument validation in the client is necessarily a good idea.  We'll have to add a special parameter for every type of validation we want to let people do (as compared to validating on the server, where they can just write whatever JS code they want), and the only benefit is that it reduces the time to first error by like 50ms.

(Also, I would guess that people will be doing validation on the server _anyway_, especially if their semantics are at all complicated.)

I would at least strongly consider making these arguments optional.

rfcs/plugins.md
```diff
@@ -0,0 +1,349 @@
...
+    - `minArgs`: minimum number of arguments accepted
+    - `maxArgs`: maximum number of arguments accepted
```
mlucy: Is there a way to specify an unlimited number of arguments?

rfcs/plugins.md
deontologician: Ok, you mean some external command. I was thinking the external process would be javascript code itself, so the plugin process manager would run it for you.

To be honest, this part isn't thought out too well. It's intended to allow worker processes that Horizon will automatically monitor and restart for you. I'm open to better proposals, the other sections got a bit more attention

rfcs/plugins.md
```diff
@@ -0,0 +1,349 @@
...
+    - `legalToChainFrom`: a list of terms that can be chained before
+      this request method. The special value `"/"` indicates the
+      method is available directly on the `horizon`
+      instance. `"collection"` indicates it's chainable from a
+      collection.
```
mlucy: Is there a way to specify that it's legal to chain from anything?

rfcs/plugins.md
```diff
@@ -0,0 +1,349 @@
...
+    - `nullable`: a boolean determining whether any arguments can be
```
mlucy: What are the semantics if this is `false`?  We error out in the client if any of the arguments are `null`?  I'm probably missing something here, because this seems like a super-specific option to be one of the first 4 we add to the interface.

rfcs/plugins.md
deontologician: These seem reasonable suggestions

rfcs/plugins.md
```diff
@@ -0,0 +1,349 @@
...
+The function should return an object which will be passed to various
+other callbacks for the plugin and contains all internal state the
+plugin needs to function. This includes keeping a reference to the
+rethinkDB connection if necessary.
+
+### Deactivate
+
+The `deactivate` plugin property must be present and contain a function that accepts the current plugin state and frees any resources the plugin is using. (Disconnect any connections, close any files, etc)
+
+## User control over plugins
```
mlucy, line 296: One of the goals here is to let users fix plugins with conflicting names, right?  We should probably let them change environment variable names too then.

rfcs/plugins.md
deontologician: Technically, yes. It can run any Node code. So while we'd probably suggest that it be side-effect free, you can make something horribly inefficient. Of course the user opts into calling it, so they can opt out if they want

rfcs/plugins.md
deontologician: I think it's reasonable that they be optional, but these client side restrictions are just the "general shape" restrictions we enforce on ast methods currently.

mlucy:
How are we going to handle errors in plugins?  Specifically:
- What happens if an exception is thrown from e.g. a request handler?
- Is there a way for plugins to report that they're experiencing internal problems and should be deactivated?
- Is there a way for plugins to log warnings and errors?
- If a plugin fails, should we automatically try to reload it, or just leave it disabled?  (Should we give plugins a way to request that they be reloaded to reset e.g. any connection state they might have that's messed up?)

rfcs/plugins.md
deontologician: Nope, and none of the built in commands allow an unlimited number either (they just use 100 as the max). This is arguably stupid, but I think it simplified some code somewhere

rfcs/plugins.md
```diff
@@ -0,0 +1,349 @@
...
+### Activate
+
+The `activate` plugin property must be present and contain a function
+that accepts an object with the following properties:
+  - `config`: config options from the config.toml section for the
+    plugin. This includes user overrides which are not specified in
+    the `config` plugin property.
+  - `env`: an object with relevant environment variables (defined in
+    [environment](#environment))
+  - `rdbConn`: a function that returns an open rethinkdb connection
```
mlucy: Is it safe to call this multiple times (e.g. if we're briefly disconnected from RethinkDB and the connection they got the first time is invalidated)?

rfcs/plugins.md
deontologician: Is that reasonable? I think in order to process anything in a reasonable way you need to know what you're chaining off of. Also, there are only a few core methods, and you can only define a terminal (like fetch or watch) so it's probably not necessary

rfcs/plugins.md
deontologician: This is sort of an artifact of the current validation scheme. Right now we have a bunch of tests that need to throw errors if you pass null. I agree this is kinda specific so we  could take it out

deontologician:
> What happens if an exception is thrown from e.g. a request handler?
- [ ] Horizon should remain stable and return an error response from a request handler, a 500 response from a route handler, and print an error message on the command line

>  Is there a way for plugins to report that they're experiencing internal problems and should be deactivated?

Open to suggestions on this one

> Is there a way for plugins to log warnings and errors?
- [ ] we should explicitly expose the logger (possibly with a log prefix) to plugins

> If a plugin fails, should we automatically try to reload it, or just leave it disabled? (Should we give plugins a way to request that they be reloaded to reset e.g. any connection state they might have that's messed up?)

This is interesting. It seems like it probably shouldn't be a plugin setting, but some general policy. Given that the plugin in question is buggy, it sort of implies the author of the plugin doesn't know where the bug is, so it's not clear whether it's ok to restart it or not. (Say it crashes before or after some system resource is allocated, it might or might not be a good idea to restart it)

deontologician:
Ok, at this point, the RFC is pretty big, if you're going to comment, please do it in Reviewable, where I can keep track of things better.

rfcs/plugins.md
deontologician: This is in general a hard thing, since we don't provide resumable changefeeds. I'd say we should expect dispose() to be called before activate is called a second time

rfcs/plugins.md
mlucy: I could imagine wanting to write e.g. a plugin that takes a query, logs it somewhere, and then runs it as normal.  (So `collection.whatever.log` or something.)  But it probably isn't a big deal since there are so few things to enumerate.

abastardi:
Review status: 0 of 1 files reviewed at latest revision, 15 unresolved discussions.

---

_[rfcs/plugins.md, line 5 [r6]](https://reviewable.io:443/reviews/rethinkdb/horizon/588#-KKb3pNHOxlRgz8VWtJg:-KKb3pNHOxlRgz8VWtJh:1445261185) ([raw file](https://github.com/rethinkdb/horizon/blob/55d3a7de9d89faa54d4c57c4227cfc12f53cf2fd/rfcs/plugins.md#L5)):_

> The plugin system I'm proposing isn't for node developers, it's for people using horizon in serverless apps who would like to get additional functionality. ...I think it would be a nice bonus to make it easy for node developers to be able to use horizon plugins to save some work, but it's a secondary or tertiary goal here.

It might help to make the above more explicit in the RFC, as that was not entirely clear (particularly as there is currently no documentation on how one could go about writing a custom backend that does the things listed in this RFC). 

It would also be helpful, perhaps as a separate RFC, to articulate proposed changes to the backend API that will (a) be needed to enable the plugin functionality, and (b) make it easier to develop custom backend functionality (via embedding Horizon rather than via plugins). For example, will any hooks be added to the request-response cycle? Will it be possible to create or add to rulesets programmatically as well as incorporate custom validation methods? Right now there is the `add_request_handler` method, but there doesn't appear to be much that enables more fine-grained control over any of the standard request handlers. The roadmap includes, "Backend -- an API/protocol to integrate custom backend code with Horizon server/client-libraries" -- it would be great to see this fleshed out further.

---

_[rfcs/plugins.md, line 16 [r6]](https://reviewable.io:443/reviews/rethinkdb/horizon/588#-KKb5g1Z-e6y7fvmEpd1:-KKb5g1_PEI8TNEGjbJd:706133259) ([raw file](https://github.com/rethinkdb/horizon/blob/55d3a7de9d89faa54d4c57c4227cfc12f53cf2fd/rfcs/plugins.md#L16)):_

> ``` Markdown
>  - Worker processes ([#368](https://github.com/rethinkdb/horizon/issues/368))
>  - Scheduled tasks ([#311](https://github.com/rethinkdb/horizon/issues/311))
> ```

It might help to explain a bit how plugins can be used to achieve some of the above, particularly things like custom authentication methods, server-side rendering, and scheduled tasks.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/588)_

deontologician:
Review status: 0 of 1 files reviewed at latest revision, 15 unresolved discussions.

---

_[rfcs/plugins.md, line 5 [r6]](https://reviewable.io:443/reviews/rethinkdb/horizon/588#-KKb3pNHOxlRgz8VWtJg:-KKkaRdoHKANF8pjzveI:-2076621130) ([raw file](https://github.com/rethinkdb/horizon/blob/55d3a7de9d89faa54d4c57c4227cfc12f53cf2fd/rfcs/plugins.md#L5)):_

> It might help to make the above more explicit in the RFC

I agree, sorry for the confusion

> It would also be helpful, perhaps as a separate RFC, to articulate proposed changes to the backend API [...]

I think we should do this once this plugin api is implemented, as well as propose a proper solution for middleware in embedded horizon ala #24, since it'll entail refactoring that @tryneus will be in the best position to make decisions on as he's writing it. Once plugins are in, it'll be easier to do some minor changes to the details to create an api that will be nice to use.

---

_[rfcs/plugins.md, line 16 [r6]](https://reviewable.io:443/reviews/rethinkdb/horizon/588#-KKb5g1Z-e6y7fvmEpd1:-KKkbgxUMRSPFulvjcAt:-169179862) ([raw file](https://github.com/rethinkdb/horizon/blob/55d3a7de9d89faa54d4c57c4227cfc12f53cf2fd/rfcs/plugins.md#L16)):_
<details><summary>

<i>Previously, abastardi wrote…</i></summary>



> It might help to explain a bit how plugins can be used to achieve some of the above, particularly things like custom authentication methods, server-side rendering, and scheduled tasks.
> </details>
> This is a good idea too, but not sure if I'll have time to flesh these out before we start writing the plugin api. It seems like it's converging pretty quickly

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/588)_

deontologician:
@mlucy @Tryneus see further updates in 599f5ea

deontologician:
I still don't think we've addressed logging or any mechanism for a plugin to declare it needs to be restarted (deactivated then reactivated)

marshall007:
Another thing I just thought of is that plugins exposing `requests` or `httpRoutes` might need a mechanism for making authenticated requests using OAuth credentials from the current user's session. For example, if users authenticate using GitHub, it's likely that the HTTP routes and WS requests you want to expose need to make authenticated calls to the GitHub API on behalf of the user.

deontologician:
@bopjesvla made some relevant comments on this proposal in #607 https://github.com/rethinkdb/horizon/pull/607#issuecomment-229484465. 

Also, I think we should perhaps provide a `plugins` metadata table where plugins can preserve their state. Thoughts @Tryneus ?

deontologician:
@marshall007 What kind of credentials would be needed? I don't think we store anything other than "this user was authenticated". Would we need #216 ?

marshall007:
@deontologician I just mean the credentials obtained from the OAuth handshake. Not the user account/profile info, but it's the same credentials you'd use to make that request too.

bopjesvla:
oh whoops

> Plugins should also be able to alter inserted documents. That'd give way for plugins to add stuff like:
> 
> Random API keys to every created user
> Document creation and modification metadata to documents in specified collections
> Estimated location based on IPs
> Right now, all data in a horizon app has to come from the untrusted client. That's simply not workable.

## 

> I agree, but I think this should be done by defining new terminals, like hz.createUserWithApiKey(...) or hz('foo').storeSpecial({...}), so that clients are clear which plugin is responsible for the behavior.

What if I want to use the API-key-adding capabilities of one plugin and the location-adding capabilities of another?

The plugin responsible for the behaviour is either the one named `horizon-plugin-add-api-key` or the one named `horizon-plugin-random` that required the user to put something like

```
[plugins.random.users.apikey]

length = 32
```

in `config.toml`. I don't think plugins surprising users by adding untraceable data will be the issue you're making it out to be.

bopjesvla:
> What if I want to use the API-key-adding capabilities of one plugin and the location-adding capabilities of another?

To be clear, this for the same document at the same insert.

deontologician:
We also need to allow plugins to add an entry to `horizon/auth_methods` which this RFC doesn't currently allow. @Tryneus might have an opinion on what we need for that

rfcs/plugins.md
```diff
@@ -0,0 +1,326 @@
...
+
+The activate function is called when initializing the plugin in the
+server. It is not invoked for `hz` commands.
+
+The `activate` plugin property must be present and contain a function
+that accepts an object with the following properties:
+  - `config`: config options from the config.toml section for the
+    plugin. This includes user overrides which are not specified in
+    the `config` plugin property.
+  - `metadata`: the
```
dalanmiller, line 156: "the"

deontologician:
Review status: 0 of 1 files reviewed at latest revision, 17 unresolved discussions.

---

_[a discussion](https://reviewable.io:443/reviews/rethinkdb/horizon/588#-KOw7jy_yazfP9HwuJlo:-KOw7jy_yazfP9HwuJlp:bs6mhod) (no related file):_
Things that need to be passed to activate in some way:
- [ ] `metadata.userTable`, a reql Table Ast object pointing to the user table
- [ ] `metadata.projectDb`, a reql Db Ast object pointing to the horizon project database

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/588)_

deontologician:
Review status: 0 of 1 files reviewed at latest revision, 17 unresolved discussions.

---

_[rfcs/plugins.md, line 156 [r8]](https://reviewable.io:443/reviews/rethinkdb/horizon/588#-KKBmnXGGtsKlY3__LEl-r8-156:-KOwDOfnIa13iYU7MODZ:bqa579e) ([raw file](https://github.com/rethinkdb/horizon/blob/599f5ea669becb66ae8bbb019c0b36768d4589e8/rfcs/plugins.md#L156)):_
<details><summary>

> "the"
> </details>
> This is supposed to be the server metadata object. I think that may be a bad idea though, it should probably be a plugin-specific object.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/588)_

Issue #564, opened 2016-06-06
Horizon can't connect to a rethinkdb proxy server


alf:
- Server version: 2.3.3, windows
- Client version: https://github.com/ericb/horizon/commit/f3a2284ff0e1ca011ca88fc22048d8f793262c9b

In the reql_connection.js, there is an assumption that the server id exists in the server_status table. This is not correct if the server we are connecting to is a proxy.

deontologician:
Nice catch

danielmewes:
@Tryneus Is this something that can be fixed easily?

Tryneus:
Fixed by PR #597.  Thanks for reporting this, @alf!

PR #597, opened 2016-06-15
Change version check to not require a non-proxy rethinkdb node

Tryneus:
Fixes issue #564.  The version check doesn't need to be so strict at the moment, because RethinkDB does not allow clusters to be made up with wildly varied binary versions.  If in the future we depend on behavior that could exist in some (but not other) nodes in a cluster, we can make a more complicated solution here.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/597)

### DIFF
--- a/server/src/reql_connection.js
+++ b/server/src/reql_connection.js
@@ -70,14 +70,12 @@ class ReqlConnection {
          logger.error(`Error on connection to RethinkDB: ${err}.`);
          retry();
        });
-       return conn.server().then((serv) =>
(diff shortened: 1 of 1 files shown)


deontologician:
Should this go out in v1.x.x or wait for 2.0?

mglukhovsky:
Since this is a bugfix, rather than a change in behavior, I think we should include it in a 1.x.x release.

deontologician:
:lgtm: merging

PR #603, opened 2016-06-16
Updates to Dockerfiles and fixes to docker-compose recipes

dalanmiller:
This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/603)

Before merging need to push `horizon` image to `rethinkdb/horizon` and `horizon-dev` to `rethinkdb/horizon-dev` on Docker Hub.


dalanmiller:
@segphault @encryptio mind reviewing this?

encryptio:
Review status: 0 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

_[.dockerignore, line 22 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUSuxI7Ze15I8mbw0s:-KKUSuxI7Ze15I8mbw0t:899674761) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/.dockerignore#L22)):_

> ```
> **/*-cert.pem
> 
> node_modules/
> ```

This is already handled by the `**/node_modules` pattern at the top of the file

---

_[Dockerfile, line 3 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUUMIfvSJUwCY2NEzX:-KKUUMIg5wSVPbMqLNmB:683055430) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile#L3)):_

> ``` Dockerfile
> FROM node:5-slim
> 
> RUN yes '' | adduser --disabled-password horizon
> ```

Multiple adjacent `RUN` commands should be combined into one to reduce the number of layers (too many does actually cause problems on some docker configurations)

e.g.

```
RUN yes '' | adduser --disabled-password horizon && \
    mkdir -p /usr/horizon && \
    ...
```

---

_[Dockerfile, line 6 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUTsCGJA8bmRlDIF0A:-KKUTsCHP1GqE42vpAe8:1736542129) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile#L6)):_

> ``` Dockerfile
> RUN mkdir -p /usr/horizon
> RUN mkdir -p /usr/app
> RUN mkdir -p /usr/certs
> ```

If the user wants to mount their certs here (which appears to be the implication of this), how does `hz` know to find them here? Does it need a `VOLUME` directive?

Or is this an extraneous `mkdir` from development cycles?

---

_[Dockerfile, line 16 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUTQz4ocP6NS0sQQF4:-KKUTQz5txMM7GYm2FKY:1689297094) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile#L16)):_

> ``` Dockerfile
> VOLUME /usr/app
> 
> # REQUIREMENTS
> ```

I'd put the requirements at the top of the file, where they look more like comments about the dockerfile in general and less like detail comments. Same in `Dockerfile.dev`. Maybe this stuff should only exist on the registry page?

---

_[Dockerfile.dev, line 1 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUUw-mkLwmwWtgp-7I:-KKUUw-ne1BaT4i1wjAl:-792925073) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile.dev#L1)):_

> ```
> FROM rethinkdb/horizon
> ```

`docker build` doesn't always fetch the tag you specify if it exists locally already (in this case, `rethinkdb/horizon:latest`); so the build for the dev dockerfile must either be done on the same machine that did the build for the production version after it did it _OR_ the production version must be pushed publicly and the production version explicitly pulled along with its `latest` tag before building. In both cases, you need to be careful to update the `latest` tag locally and/or publicly, not just the explicitly versioned tag (`latest` is the default tag but it has no special semantics.)

Alternatively, you could explicitly have the dev dockerfile be `FROM rethinkdb/horizon:1.1.3` (or whatever version is appropriate); that means most (but not all) of the mistakes you could make in the above paragraph will be replaced with docker build errors rather than silently using an old version. Of course, that comes with the cost of having to patch this dockerfile for every release, an edit which can be forgotten about.

I'm not sure what's best here. Thoughts?

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/603)_

encryptio:
Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/603)_

dalanmiller:
Review status: 2 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

_[.dockerignore, line 22 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUSuxI7Ze15I8mbw0s:-KL3WT29VtNaT6ZfU2Mr:-499146009) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/.dockerignore#L22)):_
<details><summary>

<i>Previously, encryptio (Chris Kastorff) wrote…</i></summary>



> This is already handled by the `**/node_modules` pattern at the top of the file
> </details>
> Done.

---

_[Dockerfile, line 3 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUUMIfvSJUwCY2NEzX:-KL3W2oAl-Ru5CDUWkkG:-499146009) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile#L3)):_
<details><summary>

<i>Previously, encryptio (Chris Kastorff) wrote…</i></summary>



> Multiple adjacent `RUN` commands should be combined into one to reduce the number of layers (too many does actually cause problems on some docker configurations)
> 
> e.g.
> 
> ```
> RUN yes '' | adduser --disabled-password horizon && \
>     mkdir -p /usr/horizon && \
>     ...
> ```
> 
> </details>
> Done.

---

_[Dockerfile, line 6 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUTsCGJA8bmRlDIF0A:-KL3WoTg2oEwHk0c7sVm:-1774342203) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile#L6)):_
<details><summary>

<i>Previously, encryptio (Chris Kastorff) wrote…</i></summary>



> If the user wants to mount their certs here (which appears to be the implication of this), how does `hz` know to find them here? Does it need a `VOLUME` directive?
> 
> Or is this an extraneous `mkdir` from development cycles?
> </details>
> I put this here as a standardized place where they should exist but like we kind of talked about certs and keys can be set via injecting ENV variables at runtime. 

---

_[Dockerfile, line 16 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUTQz4ocP6NS0sQQF4:-KL3X9uvQi9Og7DZ4982:2003230035) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile#L16)):_
<details><summary>

<i>Previously, encryptio (Chris Kastorff) wrote…</i></summary>



> I'd put the requirements at the top of the file, where they look more like comments about the dockerfile in general and less like detail comments. Same in `Dockerfile.dev`. Maybe this stuff should only exist on the registry page?
> </details>
> I've put them at the top for now, I feel like they are the bare minimum needed to know about running the Dockerfile so I'm going to keep them in for now. 

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/603)_

dalanmiller:
Review status: 2 of 5 files reviewed at latest revision, 5 unresolved discussions.

---

_[Dockerfile.dev, line 1 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUUw-mkLwmwWtgp-7I:-KL3YKyAYaZthrAlY6es:-1811920655) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile.dev#L1)):_
<details><summary>

<i>Previously, encryptio (Chris Kastorff) wrote…</i></summary>



> `docker build` doesn't always fetch the tag you specify if it exists locally already (in this case, `rethinkdb/horizon:latest`); so the build for the dev dockerfile must either be done on the same machine that did the build for the production version after it did it _OR_ the production version must be pushed publicly and the production version explicitly pulled along with its `latest` tag before building. In both cases, you need to be careful to update the `latest` tag locally and/or publicly, not just the explicitly versioned tag (`latest` is the default tag but it has no special semantics.)
> 
> Alternatively, you could explicitly have the dev dockerfile be `FROM rethinkdb/horizon:1.1.3` (or whatever version is appropriate); that means most (but not all) of the mistakes you could make in the above paragraph will be replaced with docker build errors rather than silently using an old version. Of course, that comes with the cost of having to patch this dockerfile for every release, an edit which can be forgotten about.
> 
> I'm not sure what's best here. Thoughts?
> </details>
> Right now `:latest` is being built automatically on repo change. Does this obviate the problem? 

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/603)_

encryptio:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

_[Dockerfile, line 3 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUUMIfvSJUwCY2NEzX:-KL3_JVpPspIdWrbxTSt:-72433095) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile#L3)):_
<details><summary>

> Done.
> </details>
> The `mkdir` calls are merged, but the user creation and mkdir are still separate.

---

_[Dockerfile.dev, line 1 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUUw-mkLwmwWtgp-7I:-KL3_aXg0FQ_BC1uGzzf:440155117) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile.dev#L1)):_
<details><summary>

> Right now `:latest` is being built automatically on repo change. Does this obviate the problem? 
> </details>
> As long as horizon finishes building and pushing before horizon-dev does, yep! :)

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/603)_

dalanmiller:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

_[Dockerfile, line 3 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUUMIfvSJUwCY2NEzX:-KL3fR0LC1ljDQEmo1hx:-972623079) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile#L3)):_
<details><summary>

<i>Previously, encryptio (Chris Kastorff) wrote…</i></summary>



> The `mkdir` calls are merged, but the user creation and mkdir are still separate.
> </details>
> I think for purposes of clarity, I don't think I should concat both lines together, one is responsible for user creation, the other is responsible for creating the base directories. 

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/603)_

dalanmiller:
Review status: 4 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

_[Dockerfile, line 3 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-KKUUMIfvSJUwCY2NEzX:-KL3qjl5MklZmr_UCxm7:-1647812371) ([raw file](https://github.com/rethinkdb/horizon/blob/ac86d7f2b3ee32b7e42e2deea9bc4f73623b5938/Dockerfile#L3)):_
<details><summary>

> I think for purposes of clarity, I don't think I should concat both lines together, one is responsible for user creation, the other is responsible for creating the base directories. 
> </details>
> All good now. We discussed and resolved civilly in person. 

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/603)_

encryptio:
:lgtm:

---

Reviewed 1 of 1 files at r3.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/603#-:-KL3sWtfU22IrutsSFUF:1417088721)_

PR #607, opened 2016-06-18
Tryneus and Mlucy's Horizon Plugins Proposal

Tryneus:
This may or may not be approved by mlucy, although we went over several of these changes together.  This was a result of a discussion with @coffeemug, looking to pare down the surface area of the proposal to something more reasonable to implement in the short term.

This proposal breaks the plugins API up into two phases, the first provides a way to add HTTP handlers and `hz` commands.  The second provides pluggable horizon websocket protocol request types and authentication methods (like `unauthenticated`, `anonymous`, and `token`).

We dropped external processes, validator extensions, client validation, and we simplified plugin configuration options.

@josh - do your _worst_.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/607)

### DIFF
--- a/rfcs/plugins.md
+++ b/rfcs/plugins.md
@@ -15,19 +15,20 @@ Plugins are intended to be flexible enough to implement things like:
  - Scheduled tasks ([#311](https://github.com/rethinkdb/horizon/issues/311))
 
 ## Things plugins can do
- - Add new configuration options
+
+Version 1:
  - Add new `hz` commands
  - Add a new http route
- - Add new helper functions available to validators
- - Spawn supervised processes
+
+Version 2:
  - Add new horizon request types useable by the client
+ - Add new horizon authentication methods useable by the client
 
 ## Things plugins can't do
 
  - hook into websocket request types it didn't define
  - respond to requests to http routes it didn't define
  - modify state of other plugins
- - override validation results
  - see configuration options for other plugins
 
 If you need to do any of these things, embedding Horizon is your best bet.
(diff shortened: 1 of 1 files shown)


deontologician:
A couple thoughts since I just have my phone, not a computer:
- overall definitely better, tightened a lot of lazy or vague stuff from previous
- I was going to drop processes as well... worker processes need more careful thinking
- metadata description truncated
- I think you should be able to define multiple request types
- I think the http routing isnt flexible enough, as it is, you cant create a rest api because you need to be able to create dynamic routes not known to exist at plugin activation time
- api for http request/response objects is missing. it just says "not express". Need to say what methods there are, how to add headers etc
- Is dropping the access to a reql connection intentional? I think that would be very convenient not to force them to create their own
- Is it possible to name the websocket request and the http route different things? I think we should allow this

abastardi:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

_[rfcs/plugins.md, line 218 [r1]](https://reviewable.io:443/reviews/rethinkdb/horizon/607#-KKb83OSPZEK8OMIDT9z:-KKb83OSPZEK8OMIDTA-:605434264) ([raw file](https://github.com/rethinkdb/horizon/blob/fd985cf6a4aa5b609c1ed8780cdddefef710e0d8/rfcs/plugins.md#L218)):_

> ``` Markdown
> 
> The handler function is a function of the format:
> `(request, context, send, done) => { ... }`
> ```

Why not make custom handler functions have the same signature as the built-in handlers (i.e., with `ruleset` and `metadata` as separate arguments)?

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/607)_

Tryneus:
@deontologician:

> worker processes need more careful thinking

We were actually of the mind that worker processes need no thinking.  Since a plugin can run arbitrary code, they can manage their own worker processes when we make the `activate` and `deactivate` calls.  There's no need to provide an API for this when Node already gives it to them.

> metadata description truncated

I haven't decided on the interface provided by the `Metadata` class, but it will probably differ a bit from the current state of things in `next`.  At a minimum, we should provide a way to get the reql connection, collections, indexes, JWT signing, and current user info.

> I think you should be able to define multiple request types

That should be possible here, and if it was omitted it is a mistake.

> I think the http routing isnt flexible enough, as it is, you cant create a rest api because you need to be able to create dynamic routes not known to exist at plugin activation time

This is a valid concern.  I'm not sure there's anything you _can't_ do a workaround with using static routes and query parameters, but I could see this being a problem if trying to match a specific API.  We could make the `done` callback from `activate` be repeatable, such that a plugin could update its definition live - but that would be slow for really large sets of paths.

> api for http request/response objects is missing. it just says "not express". Need to say what methods there are, how to add headers etc

I was imagining this would just be the default Node HTTP request/response objects.

> Is dropping the access to a reql connection intentional? I think that would be very convenient not to force them to create their own

The reql connection is available through the `Metadata`.

> Is it possible to name the websocket request and the http route different things? I think we should allow this

Yes, nothing should be preventing this.  The HTTP route is namespaced, though.

@abastardi 

> Why not make custom handler functions have the same signature as the built-in handlers (i.e., with ruleset and metadata as separate arguments)?

The idea was that we would change the built-in handlers to match this interface.  It seemed a little ugly to me to be passing in the server information as three separate objects - especially when we might want to add more later - which would break the interface.  I'm open to changing it, but it seems a bit cleaner this way.

rfcs/plugins.md
```diff
@@ -41,223 +42,140 @@ Conflicts between plugins should be easy to determine from their
...
-Example:
+Example for a plugin named `graphql`:
 
 ```js
 httpRoute: {
-    'hello-world': (req, res, state) => {
-        res.send(`Hello ${state.value}`);
-            res.end()
-    },
+  '': (req, res) => {
```
marshall007, line 142: Is there a way to specify the request method(s) or are only `GET`s going to be allowed?

deontologician:
> Yes, nothing should be preventing this. The HTTP route is namespaced, though.

What I meant is whether we allow the plugin to specify what it would like the http route namespace to be.  So it should be possible to define a request called `loggingFetch` and an http route like `/logs`. If the user has a conflict here they could remap.

>  We could make the done callback from activate be repeatable, such that a plugin could update its definition live - but that would be slow for really large sets of paths.

In the previous proposal the http route was just a prefix, so the handler could internally add or remove sub-routes at will without notifying the plugin system. I think this is preferable, since we only care about conflicts at the plugin level, and each plugin gets a non-conflicting top level route, so Horizon really doesn't need to know about those sub-routes.

rfcs/plugins.md
Tryneus, line 142: I believe the idea is that the function is passed all HTTP requests except upgrade, and it is up to the plugin to branch based on the request type.

bopjesvla:
> hook into websocket request types it didn't define

At the risk of sounding like a broken record, does this include the default database operations? I feel like the ability to hook into inserts and connects is way more important than custom HTTP routing, which seems like an uncommon requirement that can easily be handled by embedding Horizon.

deontologician:
> At the risk of sounding like a broken record, does this include the default database operations?

Yes, plugins won't be able to directly muck with `.fetch` and `.watch`, though you'll be able to define custom terminals like `hz('foo').find(3).myTerminal()` and handle the results that way. This ensures the user is explicitly stating they want to use your plugin, rather than having a plugin change the behavior of fetch or watch silently.

> I feel like the ability to hook into inserts and connects is way more important than custom HTTP routing

A lot of additional server-side options like graphQL, falcor and REST will require custom endpoints.

> an uncommon requirement that can easily be handled by embedding Horizon.

We actually want to reserve embedding for more powerful modifications that require middleware. Modifying existing database operations is exactly what embedding should be for. Plugins prioritize predictability and minimal configuration over flexibility.

bopjesvla:
> Yes, plugins won't be able to directly muck with .fetch and .watch,

I didn't mean to imply that plugins should have unlimited control. At the very minimum, I think plugins should be able to hook into requests and cancel them by throwing a special type of error, which is propagated to the client. The ability to cancel inserts, fetching/watching and connects would cover document validation, permissions and custom authentication.

> though you'll be able to define custom terminals like hz('foo').find(3).myTerminal() and handle the results that way.

That doesn't really cover the same ground, though. Also, I assume it'll look more like `hz('foo').find(3).call('myTerminal')`, since plugins can't modify the client.

> This ensures the user is explicitly stating they want to use your plugin, rather than having a plugin change the behavior of fetch or watch silently.

A canceled request would be very noisy, but in general, it's on plugin writers not to do anything crazy.

bopjesvla:
Besides canceling reqests, plugins should probably be able to alter insertions and updates to add stuff like API keys and creation date.

bopjesvla:
Plugins should also be able to alter inserted documents. That'd give way for plugins to add stuff like:
- Random API keys to every created user
- Document creation and modification metadata to documents in specified collections
- Estimated location based on IPs

Right now, all data in a horizon app has to come from the untrusted client. That's simply not workable.

deontologician:
> Plugins should also be able to alter inserted documents. [...] all data in a horizon app has to come from the untrusted client. That's simply not workable.

I agree, but I think this should be done by defining new terminals, like `hz.createUserWithApiKey(...)` or `hz('foo').storeSpecial({...})`, so that clients are clear which plugin is responsible for the behavior.

> Estimated location based on IPs

Worker processes would be a good way to do this (you probably would want to do this async, rather than blocking an insert). Plugins can get a changefeed on a given table and fire off jobs based on inserts, so it's not impossible to do with this proposal.

(Also, @bopjesvla this PR is actually merged into #588 so I'm going to move your comments there for further discussion)

PR #614, opened 2016-06-24
Allow writes on versionless rows

### LINKED ISSUE #569: Horizon should be able to update a document that has not been created by horizon
If you're reporting a bug please include the following:
- 1.1.1
- 1.1.1

If you create a document using the rethinkdb admin interface and then try to update that document using horizon, you get hit by the following error message: `socket.js:62 Uncaught Error: Object field '$hz_v$' may not be undefined`.

If you update the document using the rethinkdb admin interface and add the '$hz_v$' field with a numeric value, the problem goes away.

I would prefer it if horizon would add the field if missing instead of refusing to change the document.

Tryneus:
Fixes #569.  If we encounter a versionless row, we treat it as if it has the version `-1`, and continue normally.  This adds a few tests for versioning behavior and fixes a few small inconsistencies in write requests as well.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/614)


deontologician:
:lgtm:

---

Reviewed 9 of 9 files at r1.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/rethinkdb/horizon/614#-:-KLJRvXgR8m1XTlujUPR:1417088721)_

Issue #527, opened 2016-05-26
Add keepalives to protocol


deontologician:
This is to avoid firewalls closing websocket connections, now that we've reverted back from engine.io as the transport

I propose we add a new message type 'keepalive' that's sent at a configurable interval from the ~~server~~ client and contains nothing else. `{ "type": "keepalive" }`. The default interval should probably be ~~30 minutes or something~~ 5 minutes

danielmewes:
The client also needs to send keepalive messages, not just the server I think.
I also suggest that we make the interval significantly lower than 30 minutes. Maybe 5 minutes, or even less?

Apart from keeping routers from closing the connection, another important reason for sending keepalive messages is to detect when the connection has died. "Nothing" is worse than a realtime app that has silently stopped updating because the connection is dead. Sending something over the connection will make sure that the operating system notices if the TCP connection is no longer alive. Otherwise it could sit in a dead state over a long period and nobody would know.

deontologician:
I meant client, but typed server for some reason. So is everyone ok with 5 minutes then?

danielmewes:
Actually would it be hard to make the interval configurable? I think 5 minutes is a good default, but some people might prefer more frequent checks at the cost of slightly more server load.

deontologician:
Yeah, I was planning on making it an option in the client

babakness:
https://www.built.io/blog/2014/11/websockets-on-aws-elb/

In the comments from link:

> This issues seems to be happening because of the Idle timeout for requests in Amazon’s ELB, which is by default, 60 seconds.
> 
> You can find ‘Connection Settings’ attribute under ‘Description’ tab for ELB in AWS console. It can be modified as needed.
> 
> You have also mentioned heartbeat mechanism is in place, in such case, you’d need to ensure heartbeat frequency is less than 60 seconds to avoid losing the WebSocket connection.

Seems like a solid case for when you'd want heartbeats lower than 60 seconds.

danielmewes:
That's interesting @babakness .

This makes me think that we should consider use a lower default so users don't have to deal with this when running on AWS.

I don't have a great understanding of how much load the heartbeats are actually going to cause. Maybe we should just benchmark their overhead once this is implemented. If the overhead is very low, we might want to go for something like 50 seconds by default.

deontologician:
Implemented in #619

Issue #358, opened 2016-05-13
Provide reconnection support


deontologician:
Right now the horizon user has to implement reconnection logic themselves. Ideally, this would be handled by Horizon itself. So if the server goes away and comes back , we should restart existing feeds and re-sync etc if a disconnection occurs. This would involve not just timeout logic, but the ability for the client to re-open all existing queries and reuse the same observables etc.

Not sure what the best way to do this would be at the moment, but it's pretty important for things like mobile where connections can be spotty at times.

deontologician:
Dupe of #9

Issue #276, opened 2016-04-20
Use RxJS 5 version of websocket subject


deontologician:
Right now we have a home-rolled Websocket Subject (`client/src/socket.js`). RxJS 5 has a built in websocket module we could use, but it has a different interface from our current one. Obviously, it's better from a maintainability perspective to use the one from the library.

Then again, it may become moot if we move to something like engine.io, so we should think about this

deontologician:
Since we've gone completely 360 on engine.io since I opened this issue, this seems like a good idea again

deontologician:
Things the new websocket subject needs to do beyond the one included in rxjs 5:
- [x] Multiplex by `request_id`
- [x] Send a handshake on opening the socket
- [x] Block requests from being sent until the handshake is accepted
- [x] Send `end_subscription` messages when unsubscribing
- [x] Close the observable if the a `{type: 'error', ... }` response is received
- [x] Complete an observable that gets a `state: complete` message
- [ ] Deserialize pseudotypes, not just jsonify
- [ ] Serialize valid objects to json (including pseudotypes) before sending
- [ ] Ensure status emits:
  - [ ] open
  - [ ] ready
  - [ ] closing
  - [ ] closed

deontologician:
Current bug is that the handshake is emitted, but the connection gets closed after the handshake, so it's not still open next time you try to use the connection

deontologician:
Fixed by #619

PR #619, opened 2016-06-28
Use rxjs 5 WebSocketSubject

### LINKED ISSUE #595: Reconnecting from client doesn't cause another ready event to be sent
I'm writing something to handle reconnecting to the horizon server in case of disconnect. I listen for dissconnect using  

``` js
hz.onDisconnected(...);
```

and then retry connecting a specified number of times using

``` js
hz.connect();
```

however I never recieve another ready event after successful reconnect.
I continue to see error and disconnect events but I only ever see the one ready event.

deontologician:
Implements #276, #527 (client side at least), and should fix #595 directly.

Turns out it can't be used in a "composed" way since it does some stuff that needs to be overridden. So this is subclassing WebSocketSubject in a way that kind of throws out a bit of its functionality and implements some functionality we need like handshakes and messages that complete requests (`state: complete` should finish a multiplexed observable)

This is also laying the foundation for #358 by keeping an internal map of requests so they can be re-sent later upon reconnection. That functionality won't land in 2.0 however

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/619)

### DIFF
--- a/client/src/socket.js
+++ b/client/src/socket.js
@@ -2,6 +2,7 @@ import { AsyncSubject } from 'rxjs/AsyncSubject'
 import { BehaviorSubject } from 'rxjs/BehaviorSubject'
 import { WebSocketSubject } from 'rxjs/observable/dom/WebSocketSubject'
 import { Observable } from 'rxjs/Observable'
+import { Subscription } from 'rxjs/Subscription'
 import 'rxjs/add/observable/merge'
 import 'rxjs/add/observable/timer'
 import 'rxjs/add/operator/filter'
@@ -192,5 +193,31 @@ export class HorizonSocket extends WebSocketSubject {
         return data
       })
       .share()
+  }
+
+  // This is a hack to fix the parent's _subscribe method. Overriding
+  // _subscribe like this can be removed once
+  // https://github.com/ReactiveX/rxjs/pull/1831 is merged.
+  _subscribe(subscriber) {
+    const { source } = this
+    if (source) {
+      return source.subscribe(subscriber)
+    }
+    if (!this.socket) {
+      this._connectSocket()
+    }
+    const subscription = new Subscription()
+    subscription.add(this._output.subscribe(subscriber))
+    subscription.add(() => {
+      const { socket } = this
+      if (socket && socket.readyState === 1) {
+        if (this._output.observers.length === 0 &&
+            socket && socket.readyState === 1) {
+          socket.close()
+          this.socket = null
+        }
+      }
+    })
+    return subscription
   }
 }
--- a/client/src/socket.js
+++ b/client/src/socket.js
@@ -1,14 +1,16 @@
 import { AsyncSubject } from 'rxjs/AsyncSubject'
 import { BehaviorSubject } from 'rxjs/BehaviorSubject'
-import { Subject } from 'rxjs/Subject'
+import { WebSocketSubject } from 'rxjs/observable/dom/WebSocketSubject'
 import { Observable } from 'rxjs/Observable'
 import 'rxjs/add/observable/merge'
+import 'rxjs/add/observable/timer'
 import 'rxjs/add/operator/filter'
 import 'rxjs/add/operator/share'
+import 'rxjs/add/operator/ignoreElements'
+import 'rxjs/add/operator/concat'
+import 'rxjs/add/operator/takeWhile'
 
 import { serialize, deserialize } from './serialization.js'
-import { log } from './logging.js'
-import { WebSocket } from './shim.js'
 
 const PROTOCOL_VERSION = 'rethinkdb-horizon-v0'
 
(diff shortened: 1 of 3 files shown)


client/src/socket.js
```diff
@@ -31,14 +32,164 @@ class ProtocolError extends Error {
...
+    // it's not really a failure.
+    const keepAliveSub = Observable
+            .timer(this.keepalive * 1000, this.keepalive * 1000)
+            .map(n => ({ type: 'keepalive', n }))
+            .subscribe()
+    // Make sure keepalive is killed when the handshake is cleaned up
+    this.handshakeSub.add(keepAliveSub)
+  }
+
+  multiplex(rawRequest) {
```
marshall007: @deontologician I haven't tested any of this, but I think you might be able to re-use most of the `multiplex` logic from the base class like so:

``` js
multiplex(rawRequest) {
  const requestId = this.requestCounter++

  const subMsg = () => {
    if (this.activeRequests.size === 0) {
      // send handshake only if we're the first request to be
      // subscribed to
      this.sendHandshake()
    }

    const request = Object.assign({ request_id: requestId }, rawRequest)
    this.activeRequests.set(requestId, { request })
    return request
  }

  const unsubMsg = () => {
    this.deactivateRequest(requestId)
    return { request_id: requestId, type: 'end_subscription' }
  }

  return super.multiplex(subMsg, unsubMsg, msg => msg.request_id === requestId)
}
```

deontologician: This is true, but once I trimmed out the generality I didn't need (arbitrary filter function, tryCatch around stuff I know can't throw an exception) the function was about the same size, so I just reimplemented it and eliminated some extra function calls and closures

marshall007: They are intentionally wrapping the `tryCatch` in a closure because that's a fairly well-known workaround for ensuring the calling function can still be optimized by V8, meaning it has minimal performance impact at runtime (`bluebird` uses the same trick). I would imagine bytes saved + their fancy compiler tricks is actually better in terms of performance than reducing closures here.

The real advantage is that it's easier to read and thus make changes. I think we probably can/should do the same thing in `sendHandshake()` as well.

marshall007: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers

deontologician: Ok, I am convinced! I will keep my hand out of optimization efforts. It's probably best to re-use as much of the underlying class anyway

client/src/socket.js
```diff
@@ -136,84 +138,77 @@ class HorizonSocket extends WebSocketSubject {
...
+        error: err => subscriber.error(err),
         complete: () => this.next(request), // send request on
-                                            // handshake completion
+        // handshake completion
       })
 
-      // Now we subscribe to messages from the server with our
-      // requestId.
-      const subscription = this.subscribe({
+      this.handshake.ignoreElements().concat(
```
marshall007: @deontologician might be able to avoid creating the inner observer (`subscriber`) here:

``` js
return this.handshake.ignoreElements().concat(
  super.multiplex(subMsg, unsubMsg, res => res.request_id === requestId)
)
.takeUntil(res => res.state === 'complete')
.map(res => {
  if (res.error !== undefined) {
    // This is an error just for this request, not the
    // entire connection
    throw new ProtocolError(res.error, res.error_code)
  }
  if (res.data !== undefined) {
    // Forward on the response if there's a data field
    return res
  }
  if (res.state === 'synced') {
    // Create a little dummy object for sync notifications
    return { type: 'state', state: 'synced' }
  }
})
```

In fact, the body of the `.map()` could be moved to `resultSelector`. `throw`ing from there will propagate as an error through the observable chains just fine. I think doing so would also allow you to remove that same logic from the `handshakeSub` since the responses coming out of `super.multiplex` would already be formatted or thrown as errors.

deontologician: Ok, I think I can incorporate this

deontologician: Ended up needing to emit an extra event on `complete`, since takeWhile doesn't return the matching event

deontologician: also `concatMap`, since sometimes we need to emit multiple events for a single message from the server

marshall007: 👍 looks really good!

deontologician:
So, blargle. The subMsg wants a synchronous result, and we need to wait for the handshake to complete before emitting the subscription. And it _must_ have a subscription message, multiplex will send `null` or `undefined` over the socket if you refuse to set one (which for us will be a protocol error). It would be ideal if subMsg and unsubMsg could be provided by observables, since then we could wait for the handshake. As it is the only option is reimplementing multiplex

deontologician:
Ok, actually nevermind, it looks like concating the multiplex observable onto the handshake suffices to prevent an early request

deontologician:
This is still pending some fixes that are in the master branch of rxjs5 but haven't been merged yet (specifically https://github.com/ReactiveX/rxjs/pull/1790)

deontologician:
Ok, most recent things I've run into:
1. [x] horizon-core.js doesn't create a Horizon global
2. [x] Handshake seems to be firing off more than once per websocket connection.
3. [x] Tests still failing, though this may be a result of 2

The "It works!" page test is going though. So this is in the final stretch

alf:
The fixes in rxjs seems to be included in beta10 which is out.

deontologician:
All tests are now passing. Going to clean up the packaging situation now (hopefully simple...)

deontologician:
Also thanks to @marshall007 for a ton of help, this has been unexpectedly frustrating

marshall007:
@deontologician congrats! No problem! I think we learned a lot in the process... although, if asked, I'm still not sure I could tell you anything about how `WebSocketSubject` is supposed to work, haha.

deontologician:
Ok, I think this is ready to go (assuming circleci doesn't complain). I'm going to do the build system changes in another PR since this one is already pretty gnarly.

deontologician:
@dalanmiller or @Tryneus interested in reviewing?

deontologician:
Ok! It turns out I'm not crazy, I just forgot I needed to fix the underlying WebSocketSubject this whole this is based on. There's a PR I opened in upstream: https://github.com/ReactiveX/rxjs/pull/1831

Once that's merged and released, we can bump the rxjs version on this library and merge this PR

deontologician:
Since ReactiveX/rxjs#1831 may take some time to get merged, I added a hack to override `_subscribe` in the meantime. This now passes all tests and should be good to merge.

Issue #552, opened 2016-06-03
Remove and removeAll permissions broken


deontologician:
According to a user in slack, the following permission doesn't work:

``` toml
[groups.authenticated.rules.remove_own_messages]
template = "collection('todos').removeAll({owner: userId()})"
```

He also mentioned permissions for `remove` not working until a validator function was also added.

deontologician:
@segphault ran into this today. This is basically expected behavior currently, we should probably throw an error if you create a template on the argument to remove or removeAll since they're not related to indexes

deontologician:
Fixing this won't be a small bugfix, but a redesign of how the write-methods work. We should probably have a new command for remove/removeAll style queries that allows a search to be specified. i.e. "Find documents with this value in this field, remove them all". Then the templates would be useful for these.

marshall007:
@deontologician instead of adding a new term, do you think we could just support chaining `.remove()` onto `.find(...)` and `.findAll(...)`? We'd have to update the protocol to support "sub-queries" on write operations, but doing so would allow us to support any other terms too.

``` js
horizon('foo').findAll({ bar: 'baz' }).limit(5).remove()
```

``` json
{
  "request_id": 1,
  "type": "remove",
  "options": {
    "collection": "foo",
    "query": {
      "find_all": [ { "foo": "bar" } ],
      "limit": 5
    }
  }
}
```

deontologician:
This is a reasonable solution, we were basically 60/40 in favor of a new
term since it would reduce the amount of polymorphism on remove (it
currently accepts both objects containing an id field, and bare ids). But
using remove again isn't necessarily a bad choice.  ping @mlucy

marshall007:
I think being able to chain `above/below`, `limit`, etc before `.remove()` makes this the obvious choice over a new term. Consider use-cases like paging and wanting to delete the entire current "page" of results.

**[edit]:** we could also eliminate polymorphism by only allowing the zero-argument `.find(...).remove()` and `.findAll(...).remove()` alternatives if that's a concern (and maybe even dropping `.removeAll(...)` entirely).

deontologician:
Oh, yeah I definitely think chaining is the way to go, just whether you should be able to tack on `.remove()` or `.delete()` at the end. It might be less confusing since `.remove()` at the end would be replacing `.fetch()`

deontologician:
[...]

PR #624, opened 2016-06-30
findAll auth doesn't allow for specifying keys

BBB:
The auth template `collection("test").findAll({ owner: userId() }).fetch()` fails for a request like `{"request_id":5,"type":"query","options":{"collection":"test","find_all":[{"owner":"123"}]}}`.

I've included a failing test for this. 

Is this a bug, has the feature not been implemented yet (https://github.com/rethinkdb/horizon/blob/6d61901/server/test/permissions.js#L65 seems to suggest this). Or am I just doing this incorrectly?

I think this is related to https://github.com/rethinkdb/horizon/issues/552

I'm happy to work on this feature if no one else is. Even if it's just to throw an error when a template like this is specified.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/624)

### DIFF
--- a/server/test/permissions.js
+++ b/server/test/permissions.js
@@ -97,6 +97,30 @@ describe('Permissions', () => {
       assert(rule.is_match(make_request('query', 'test', { find_all: [ { } ], fake: 'baz' }), context));
     });
 
+    it('single key in findAll', () => {
+      const rule = new Rule('foo', { template: 'collection("test").findAll({ owner: userId() }).fetch()' });
+      assert(rule.is_valid());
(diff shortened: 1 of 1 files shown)


deontologician:
ping @Tryneus

Tryneus:
Merging this and looking into the failing test.

Tryneus:
The test appears to be failing because the protocol uses `find_all` instead of `findAll`.  After changing this the tests pass.  Fixed in commit 4cdf47b.  Don't worry about the typo, and thanks for submitting this PR, @BBB, we can always use more code coverage!

Issue #513, opened 2016-05-26
Retry writes if they fail due to another write intervening


deontologician:
If the user doesn't provide a $hz_v$ field in the document to write, the server should retry the write on version check failures up to some timeout.

If the user provides an $hz_v$ value in their document, and the version check fails, the write should fail and not retry.

deontologician:
Fixed by #636

PR #636, opened 2016-07-08
Write retries with timeout (server side)

Tryneus:
Fixes issue #513.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/636)

### DIFF
--- a/server/src/endpoint/writes.js
+++ b/server/src/endpoint/writes.js
@@ -30,7 +30,8 @@ const make_write_response = (data) => {
 //   null means no timeout
 // pre_validate -> function (rows):
 //   rows: all pending rows
-//   return: a (promise of an) array of info for those rows (which will be passed to the validate step)
+//   return: an array or the promise of an array of info for those rows
+//           (which will be passed to the validate callback)
 // validate_row -> function (row, info):
 //   row: The row from the original query
 //   info: The info returned by the pre_validate step for this row
@@ -39,13 +40,7 @@ const make_write_response = (data) => {
 //   rows: all pending rows
 //   return: a (promise of a) ReQL write result object
 const retry_loop = (original_rows, ruleset, timeout, pre_validate, validate_row, do_write) => {
-  const response_data = Array(original_rows.length).fill(null);
-
-  // Save original version fields, which may get clobbered, so we don't have to copy everything
-  let row_data = original_rows.map((row, index) => ({ row, index, version: row[hz_v] }));
-  let deadline;
-
-  const iterate = () => {
+  const iterate = (row_data, response_data, deadline) => {
     if (row_data.length === 0) {
       return response_data;
     } else if (timeout !== null) {
@@ -118,14 +113,14 @@ const retry_loop = (original_rows, ruleset, timeout, pre_validate, validate_row,
         }
       });
 
-      row_data = retry_rows;
-
       // Recurse, after which it will decide if there is more work to be done
-      return iterate();
+      return iterate(retry_rows, response_data, deadline);
     });
   };
 
-  return iterate().then(make_write_response);
+  return iterate(original_rows.map((row, index) => ({ row, index, version: row[hz_v] })),
+                 Array(original_rows.length).fill(null),
+                 null).then(make_write_response);
 };
 
 const validate_old_row_optional = (original, old_row, new_row, ruleset) => {
--- a/server/src/endpoint/writes.js
+++ b/server/src/endpoint/writes.js
@@ -128,6 +128,42 @@ const retry_loop = (original_rows, ruleset, timeout, pre_validate, validate_row,
   return iterate().then(make_write_response);
 };
 
+const validate_old_row_optional = (original, old_row, new_row, ruleset) => {
+  const expected_version = original[hz_v];
+  if (expected_version !== undefined &&
+      (!old_row || expected_version !== old_row[hz_v])) {
+    return new Error(invalidated_msg);
+  } else if (!ruleset.validate(context, old_row, new_row)) {
+    return new Error(unauthorized_msg);
+  }
+
+  if (old_row) {
+    const old_version = old_row[hz_v];
+    if (expected_version === undefined) {
+      original[hz_v] = old_version === undefined ? -1 : old_version;
+    }
+  }
+};
+
+const validate_old_row_required = (original, old_row, new_row, ruleset) => {
+  if (old_row === null) {
+    return new Error(missing_msg);
+  }
+
+  const old_version = old_row[hz_v];
+  const expected_version = original[hz_v];
+  if (expected_version !== undefined &&
+      expected_version !== old_version) {
+    return new Error(invalidated_msg);
+  } else if (!ruleset.validate(context, old_row, new_row)) {
+    return new Error(unauthorized_msg);
+  }
+
+  if (expected_version === undefined) {
+    original[hz_v] = old_version === undefined ? -1 : old_version;
+  }
+};
+
 module.exports = {
   invalidated_msg,
   missing_msg,
@@ -137,4 +173,6 @@ module.exports = {
   version_field: hz_v,
   apply_version,
   retry_loop,
+  validate_old_row_required,
+  validate_old_row_optional,
 };
--- a/server/src/endpoint/writes.js
+++ b/server/src/endpoint/writes.js
@@ -11,8 +11,8 @@ const missing_msg = 'The document was missing.';
 const timeout_msg = 'Operation timed out.';
 const unauthorized_msg = 'Operation not permitted.';
 
-const version_field = '$hz_v$';
-const apply_version = (row, new_version) => row.merge(r.object(version_field, new_version));
+const hz_v = '$hz_v$';
+const apply_version = (row, new_version) => row.merge(r.object(hz_v, new_version));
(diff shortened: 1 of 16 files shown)


deontologician:
ping @mlucy

server/src/endpoint/writes.js
```diff
@@ -6,46 +6,135 @@ const r = require('rethinkdb');
...
+};
+
+// This function returns a Promise that resolves to an array of responses - one for each row in
+//  `original_rows`, or rejects with an appropriate error.
+// timeout -> integer
+//   minimum number of milliseconds before giving up on retrying writes
+//   null means no timeout
+// pre_validate -> function (rows):
+//   rows: all pending rows
+//   return: a (promise of an) array of info for those rows (which will be passed to the validate step)
```
mlucy: This is a little unclear, it makes it sound like it has to be a promise.

mlucy:
Am I correct that most of the first part of this change is just rejiggering the existing logic to use the new `retry_loop` interface?  Did anything change enough that it needs particular attention?

server/src/endpoint/writes.js
```diff
@@ -6,46 +6,135 @@ const r = require('rethinkdb');
...
+        if (data.version === undefined) {
+          delete data.row[hz_v];
         } else {
-          data[i] = { error: res.error };
-        }
-      } else if (res.new_val === null) {
-        data[i] = { id: res.old_val.id };
-        if (res.old_val[version_field] !== undefined) {
-          data[i][version_field] = res.old_val[version_field];
+          data.row[hz_v] = data.version;
```
mlucy, line 71: I'm a little confused by these last few lines.  Didn't we just stat `data.version` to `data.row[hz_v]` in the `map` above?  I don't see us resetting it anywhere.

server/src/endpoint/writes.js
```diff
@@ -6,46 +6,135 @@ const r = require('rethinkdb');
...
+      check(changes.length === row_data.length);
+
+      // Remove successful writes and invalidated writes that had an initial version
+      const retry_rows = [ ];
+      row_data.forEach((data, index) => {
+        const res = changes[index];
+        if (res.error !== undefined) {
+          if (res.error.indexOf('Duplicate primary key') === 0) {
+            response_data[data.index] = { error: 'The document already exists.' };
+          } else if (res.error.indexOf(invalidated_msg) === 0 &&
```
mlucy, line 109: Why is 'Duplicate primary key'` hardcoded when this isn't?

server/src/endpoint/writes.js
```diff
@@ -6,46 +6,135 @@ const r = require('rethinkdb');
...
+                     data.version === undefined) {
+            retry_rows.push(data);
+          } else {
+            response_data[data.index] = { error: res.error };
+          }
+        } else if (res.new_val === null) {
+          response_data[data.index] = { id: res.old_val.id, [hz_v]: res.old_val[hz_v] };
```
mlucy, line 116: What's the semantic meaning of returning the `hz_v` of the old value here?  It seems weird that I could do an insert, get a version, then do a delete and get the same version even though it's for a later write.

mlucy, line 116: I'm not sure exactly how these versions are used, but it seems like you'd want a special version to represent "is deleted".

server/src/endpoint/writes.js
```diff
@@ -6,46 +6,135 @@ const r = require('rethinkdb');
...
+        } else {
+          response_data[data.index] = { id: res.new_val.id, [hz_v]: res.new_val[hz_v] };
+        }
+      });
+
+      row_data = retry_rows;
+
+      // Recurse, after which it will decide if there is more work to be done
+      return iterate();
```
mlucy: I'm not sure if I'm reviewing this for style or just correctness, but you seem to be going out of your way to introduce global state here.  I would strongly consider writing this so that `iterate` takes the data it's operating on -- so instead of setting a global `row_data` here, you'd write `return iterate(retry_rows)` and `iterate` would be written `(row_data) => ...`.  (I also wouldn't name it "iterate".)

(I'd also consider making `iterate` actually return something useful rather than writing to an external `response_data` variable, but that's a slightly nontrivial change which might not be worth the effort even if it's cleaner.)

server/src/permissions/rule.js
```diff
@@ -40,6 +40,16 @@ class Ruleset {
     this._rules = rules;
   }
 
+  validation_required() {
+    for (const rule of this._rules) {
+      if (!rule._validator) {
+        return false;
```
mlucy, line 46: Is this right?  Are the semantics that only one rule needs to match?

mlucy:
The writes for this seem a tiny bit sketchy -- they're only really testing one code path, and they don't seem to be testing the timeout functionality at all.  I think we might actually need an integration test for this where there's a bunch of write contention for one document.

server/src/endpoint/writes.js
Tryneus, line 109: Because 'Duplicate primary key' is generated by RethinkDB and only used here, while `invalidated_msg` is used in many places and originates from us.

server/src/endpoint/writes.js
Tryneus, line 116: The user knows what kind of operation they issued.  If it's a remove, we return the version of the document that was removed.  For all other writes we return the new version.

server/src/endpoint/writes.js
Tryneus: `iterate` does return something useful, a promise which will eventually resolve to `response_data` once all rows have succeeded or failed (due to something other than invalidation by another write), or the timeout has expired.

I like the name `iterate` because it's a fair description of its task - each call represents an iteration of the retry loop - which is the name of the function it is defined inside.  I'm open to suggestions for another name if you have any.

I'll look into lowering the amount of local global state.

server/src/permissions/rule.js
Tryneus, line 46: Since the set of rules is considered a whitelist, if the query matches at least one template and passes the associated validator, the request is allowed.  Therefore if a matching rule does not have a validator, validation is not needed.

Tryneus:
The write tests were meant to exercise the retry loop, which was the most complicated piece of the change.  I am not very concerned about the timeout logic itself, as it's fairly simple, but it would be easy enough to test a timeout of `0` without an integration test.  Beyond that, it would take some effort.

server/src/endpoint/writes.js
Tryneus, line 71: The validator callback for several write types may modify the expected version of the write based on what was found in the pre-validator callback.  This could have been done differently with a deep copy of the data, but since I don't want to introduce performance penalties for large rows, I avoided that.

server/src/endpoint/writes.js
Tryneus: Fixed in the latest commits.

server/src/endpoint/writes.js
Tryneus: I moved `row_data`, `response_data`, and `timeout` down one scope to be parameters to the `iterate` call.  They are still global in the context of a single call to `iterate`, because chaining them from promise to promise is rather difficult, especially if one promise is from a ReQL query (as I'm sure you've encountered).  So, not much has changed, but I did what I could.

Tryneus:
Added tests for a zero timeout, which actually did uncover a bug in completely unrelated code, so +1 on that.

server/src/endpoint/writes.js
mlucy, line 71: I'm still a little confused.  Why would the validator set the version on the row?  What does it mean if it sets it to a particular value?  And why is it OK to discard that value here?

(It might be nice to have that in a comment, actually, since it's kind of non-obvious why the validators would change this and what they change it to semantically.)

server/src/endpoint/writes.js
mlucy, line 116: We let people issue writes that only get applied if a document is a particular version, right?  How do I issue a write saying "apply this if the document doesn't exist"?

(Also, what if I want to write generic code that takes several Horizon write operations and pipelines them, using the return value of each as the input version for the next one so we abort if someone else writes in the middle?  That code would then have to examine the type of the write it's doing to figure out whether that particular write returns a version representing the current state of things or a version representing the previous state of things.)

(This is totally outside the scope of this review, by the way; I'm just curious.)

mlucy:
This looks basically good to me, feel free to merge it whenever.

server/src/endpoint/writes.js
Tryneus, line 116: If you only want a write to apply if the document doesn't exist, you use an `insert` request.

If you want generic chaining, you can wrap the remove response to give you the format you expect.  In that case, it is a little annoying to have the behaviors be different, but I think it is better to give developers the ability to determine what exactly was removed.

server/src/endpoint/writes.js
Tryneus, line 71: The validation callback only approves moving from one specific version to another.  If the write did not specify a specific version to apply to, we need to make sure the final ReQL write will only go through if the version has not changed since validation.  So in the validator callback, if no specific version has been chosen, we lock it to the version of the old row (this is what gives us an invalidated error if the row has changed by the time we get back to the data layer).

I'll add a comment for this before merging.

Issue #634, opened 2016-07-07
Interacting with Horizon's observables is an enormous pain.


mlucy:
The single biggest usability problem I've had getting up and running is working with Horizon's observables.  I had three major problems:
1. It isn't obvious what's defined and what isn't.
   - This was extremely frustrating.  We just link to outside documentation on observables, which is reasonable, but has the unfortunate side-effect that I couldn't actually run most examples because we happened not to include some of the operations.
2. It isn't obvious how to add new operations.
   - We have no documentation on our site about how to do this.
   - I _still_ don't know how to do it in the case where you're just writing a script instead of using a build system.
   - It ended up being easier to just download and rebuild Horizon so I could get a client with the commands I needed pre-included.
3. It isn't obvious how to access Horizon's `Observer` class.
   - You need this because some observer commands check that their arguments are from the same class, so you can't e.g. write `hz.currentUser().watch().takeUntil(Rx.Observable.timer(5000))`.
   - I literally ended up writing `Observable = hz.currentUser().watch().constructor` in my code.

I think we can fix the first one by just providing all of `rx-lite` out of the box.  I think the gains in terms of convenience far outweigh the costs in terms of code size.

@Tryneus suggested that we could maybe fix (2) and (3) at the same time by letting users pass us the Observable class they want to use, but I don't know if that's reasonable or not.  (We could also just fix (3) by exporting our `Observable` class somewhere and documenting it.)

alf:
Could you describe how to add operators when using a build system? Would be nice with a workaround until this issue is fixed.

deontologician:
Right now, the easiest way if you have webpack or browserify is to import the `rxjs/add/*` modules, which will monkeypatch Observable. So, for example, to add `reduce`, you'd do:

``` js
import { Observable } from 'rxjs/Observable'
import 'rxjs/add/operator/reduce'
// or in es5:
var Observable = require('rxjs/Observable').Observable
require('rxjs/add/operator/reduce')

// use it like:
myObservable.reduce(...)
```

Alternately, you can grab the bare operators and use them with `.call` or the [function-bind syntax](https://babeljs.io/blog/2015/05/14/function-bind) from babel. The main benefit of this is that it doesn't monkeypatch the global observable class.

``` js
import { Observable } from 'rxjs/Observable'
import { reduce } from 'rxjs/operators/reduce'

// use it like:
reduce.call(myObservable, other, args, go, here)
// or with the :: operator from babel:
[...]

PR #638, opened 2016-07-14
Make horizon observables not so bad

deontologician:
Related to #634 

This sets out to:
- [X] Patch in all observable methods in the default kitchen-sink horizon.js mode
- [x] Export Rx as a global in kitchen-sink mode
- [x] Don't bundle rx in horizon-core.js (currently it gets bundled anyway :disappointed: )
- [x] Don't break node.js in doing this

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/638)


deontologician:
Ok, looks like we're good to go. Now when you use `horizon.js` you'll get Rx as a global, and you'll get all Observable methods bundled. If you use `horizon-core.js`, you get no Rx and have to bring your own

mlucy:
Looks good to me.

Issue #639, opened 2016-07-14
Embedding Horizon server should immediately fail when trying to make the first connection to RethinkDB


dalanmiller:
At the moment, embedding Horizon server will infinitely try to connect to RethinkDB and log no errors when it is unable to connect nor eventually fail after N reconnect attempts. 

We need to add two things:
1. In the `Server` class constructor it should immediately throw an error if the first connection attempt fails. 
2. We should log failures to connect. 

Thank you to @casche for reporting this. Note: @Tryneus.

casche:
thanks 👍 @dalanmiller for helping out!

Tryneus:
So, I'm not entirely convinced this is possible or a good idea.  First of all, the `Server` constructor cannot throw when the first connection attempt fails because that happens asynchronously.  I think the best way to handle this is to get the Promise from `server.ready()`, and timer-supervise it.  We _could_ make this promise error if there is a connectivity issue on the first attempt (but not a problem with table availability or metadata correctness, I suppose?) if you guys prefer.

The server not logging on failure to connect appears to be a bug (the messages are classified as debug, while they should be error).

deontologician:
I remember changing the logs to debug since they would not stop spamming while they were waiting for tables to be ready. Ideally, a connectivity error and a table-not-ready error should be disambiguated, with the former failing immediately and logging an error, and the latter maybe not logging anything until it times out

casche:
How does Horizon behave if the DB connection is ever severed? 

I am now using docker-compose to start up my horizon app and RethinkDB. 

```
web:
  build: .
  links:
    - db
  ports:
    - "8080:80"
db:
  image: rethinkdb
  ports:
    - "27015:27015"
```

Docker compose cannot guarantee the [starting order of these services](https://docs.docker.com/compose/startup-order/) (without a kludge). So far the application starts first and the database second. 

```
Creating beauty_db_1
Creating beauty_web_1
Attaching to beauty_db_1, beauty_web_1
web_1  | 
web_1  | > lightbrew-beauty@0.0.1 prestart /usr/src/app
web_1  | > babel-node tools/startMessage.js
web_1  | 
web_1  | Starting app in dev mode...
web_1  | 
web_1  | > lightbrew-beauty@0.0.1 start /usr/src/app
web_1  | > node tools/distServer.js
web_1  | 
web_1  | info: Connecting to RethinkDB: localhost:28015
db_1   | Recursively removing directory /data/rethinkdb_data/tmp
[...]

PR #644, opened 2016-07-16
Wait for internal db on startup and always print error message on reql connection failure

Tryneus:
As part of the fix for #639, these changes should make it more obvious when there is a connection problem to the database.  Reorganized some of the code in the `Metadata` constructor to be a more linear promise chain as part of this.

This does not address all aspects of #639, as these is still no mechanism for passing errors to the user of the `Server` object when there is connectivity problems with the database - it is only reported through the logger.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/644)

### DIFF
--- a/server/src/metadata/metadata.js
+++ b/server/src/metadata/metadata.js
@@ -65,9 +65,34 @@ class Metadata {
     this._group_feed = null;
     this._index_feed = null;
 
-    const make_feeds = () => {
-      logger.debug('running metadata sync');
-      const groups_ready =
+    this._ready_promise = Promise.resolve().then(() => {
+      logger.debug('checking for internal db/tables');
+      if (this._auto_create_collection) {
+        return initialize_metadata_reql(r, this._internal_db, this._db).run(this._conn)
+      } else {
+        return r.expr([ this._db, this._internal_db ])
+                 .concatMap((db) => r.branch(r.dbList().contains(db), [], [ db ]))
+                 .run(this._conn)
+                 .then((missing_dbs) => {
+          if (missing_dbs.length > 0) {
+            let err_msg;
+            if (missing_dbs.length === 1) {
+              err_msg = `The database ${missing_dbs[0]} does not exist.`;
+            } else {
+              err_msg = `The databases ${missing_dbs.join(' and ')} do not exist.`;
+            }
+            throw new Error(err_msg + '  Run `hz set-schema` to initialize the database, ' +
+                            'then start the Horizon server.');
+          }
+        });
+      }
+    }).then(() => {
+      logger.debug('waiting for internal db');
+      return r.db(this._internal_db).wait({ timeout: 30 }).run(this._conn)
+    }).then(() => {
+      logger.debug('syncing metadata changefeeds');
+      return Promise.all([
+        // Groups changefeed
         r.db(this._internal_db)
           .table('groups')
           .changes({ squash: true,
@@ -102,8 +127,9 @@ class Metadata {
                 }
               }).catch(reject);
             });
-          });
-      const collections_ready =
+          }),
+
+        // Collections changefeed
         r.db(this._internal_db)
           .table('collections')
           .changes({ squash: false,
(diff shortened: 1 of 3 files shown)


casche:
tl;dr;

Hey @Tryneus thanks again for looking at this 👍. Additionally, it would be helpful in this case to also receive positive feedback when Horizon connects on startup in embedded mode. 

for example:

```
info: Connecting to RethinkDB: localhost:28015
```

then

```
info: Connected to RethinkDB: localhost:28015

```

In my architecture, I have the horizon application and rethinkdb in their own containers. When they start I only receive the feedback that it is connecting. I never know it 'connected'. 

## 

Long version 😆 

server.js

``` javascript
const app = express();

app.get('*', function(req, res) {
  res.sendFile(path.join( __dirname, '../src/index.html'));
});

const httpServer = app.listen(appConfig.port, function(err) {
  if (err) {
    console.log(err);
  } else {
    open(`http://localhost:${appConfig.port}`);
  }
});

const horizonServer = horizon(httpServer, {
   auto_create_collection: true,
   auto_create_index: true,
   project_name: appConfig.title,
   permissions: false,
   auth: {
     allow_anonymous: true,
     allow_unauthenticated: true,
     token_secret: appConfig.token_secret
   }
 });

 export default httpServer;
```

docker-compose.yml

``` yml
version: "2"
services:
  web:
    build: .
    depends_on:
      - db
    ports:
      - "80:8080"
  db:
    image: rethinkdb
    ports:
      - "28015:28015"
```

As you might expect the docker container just runs `npm start` which runs `node server.js`. In this setup I never know if the connection occurred:

``` bash
$ docker-compose up
Creating network "beauty_default" with the default driver
Recreating beauty_db_1
Recreating beauty_web_1
Attaching to beauty_db_1, beauty_web_1
web_1  | 
web_1  | > lightbrew-beauty@0.0.1 prestart /usr/src/app
web_1  | > babel-node tools/startMessage.js
web_1  | 
db_1   | Running rethinkdb 2.3.4~0jessie (GCC 4.9.2)...
db_1   | Running on Linux 4.4.12-boot2docker x86_64
db_1   | Loading data from directory /data/rethinkdb_data
db_1   | warn: Cache size does not leave much memory for server and query overhead (available memory: 680 MB).
db_1   | warn: Cache size is very low and may impact performance.
db_1   | Listening for intracluster connections on port 29015
db_1   | Listening for client driver connections on port 28015
db_1   | Listening for administrative HTTP connections on port 8080
web_1  | Starting app in dev mode...
web_1  | 
web_1  | > lightbrew-beauty@0.0.1 start /usr/src/app
web_1  | > node tools/distServer.js
web_1  | 
db_1   | Listening on cluster addresses: 127.0.0.1, 172.18.0.2, ::1, fe80::42:acff:fe12:2%39
db_1   | Listening on driver addresses: 127.0.0.1, 172.18.0.2, ::1, fe80::42:acff:fe12:2%39
web_1  | info: Connecting to RethinkDB: localhost:28015
db_1   | Listening on http addresses: 127.0.0.1, 172.18.0.2, ::1, fe80::42:acff:fe12:2%39
db_1   | Server ready, "2a361cf80bfd_wur" 51785a58-556a-45d4-9277-1b54b430fdc8
```

server/src/metadata/metadata.js
```diff
@@ -201,35 +228,8 @@ class Metadata {
                 }
               }).catch(reject);
             });
-          });
-      return Promise.all([ groups_ready, collections_ready, indexes_ready ]);
```
deontologician, line 205: This syntax of giving them intermediate variables with meaningful names seems better than the new version

deontologician:
Other than the one comment about clarity, lgtm :+1:

Tryneus:
@deontologician - I added back in the changefeed variable names for clarity.

@casche - we already had a log statement for that (`Metadata synced with database, ready for traffic.`), but I changed it to `Connection to RethinkDB ready: user @ host:port` similar to your suggestion.  If you want to check for readiness programmatically, you should use the `server.ready()` promise.  This will be reset to a new Promise every time we lose the connection to the database - `server.is_ready()` should indicate if that is necessary.

casche:
@Tryneus 👍 ⭐ :shipit:

Issue #583, opened 2016-06-11
Create Yeoman generators for horizon and for different frontend frameworks


deontologician:
There was some discussion in #405 about having a scaffolding function in the hz tool, but we decided it was probably better to just have some template generators for an existing tool like Yeoman.

We should probably make a new directory for them like `yeoman/` and put different npm packages in there, so we can keep versions in sync

PR #647, opened 2016-07-18
yeoman generators 

deadcee:
This is casting #583.
This includes a angular2 and a schema generator. All help accepted.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/647)


deontologician:
Hi @deadcee, thanks for the PR!

This is more new code than we can take on maintenance for at the moment. Would you be willing to create an independent package and maintain it? If not, we could leave this PR open for a while until we're ready to add generators and really maintain them.

deadcee:
I think leaving this pull request open would be the best option.

Issue #444, opened 2016-05-19
Allow queries based on nested fields


deontologician:
Right now we only build indices/allow querying based on top level fields. This is a fairly arbitrary restriction, we should see if we can support nested fields in an elegant way that doesn't make the api too complicated

marshall007:
I started looking into this. Currently we generate index names by joining on `_` and thus have to escape `_` and `\` characters. Inevitably, in order to support nested fields we are going to have to come up with an additional escaped character (probably `.`).

Instead of doing that, I think as a first step we should consider generating index names by serializing the fields to JSON instead of joining on `_`.

| Fields | Current | Proposed |
| :-- | :-- | :-- |
| `[ 'foo', 'bar' ]` | `foo_bar` | `{"foo":true,"bar":true}` |
| `[ 'foo_bar', 'baz' ]` | `foo\_bar_baz` | `{"foo_bar":true,"baz":true}` |
| `[ 'foo.bar', 'baz' ]` | N/A | `{"foo":{"bar":true},"baz":true}` |

This makes the secondary index names much easier to consume (could even be done in ReQL!), helps us avoid future corner cases with escaping field names, and provides a bit more flexibility over `_` delimited strings.

marshall007:
We would either need to use something like [`json-stable-stringify`](https://github.com/substack/json-stable-stringify) or go ahead and make key ordering deterministic in RethinkDB as was briefly discussed in [rethinkdb/rethinkdb#4636 (comments)](https://github.com/rethinkdb/rethinkdb/pull/4636#issuecomment-135575558).

deontologician:
Ah, that's pretty interesting

Tryneus:
@marshall007 - I don't think we need deterministic key order, rather we need to know the order that the fields go into the index (this is important for matching queries to indexes on range gets).  Therefore, I propose a slight modification to your suggestion:

| Fields | Current | Proposed |
| :-- | :-- | :-- |
| `[ 'foo', 'bar' ]` | `foo_bar` | `{"foo":0,"bar":1}` |
| `[ 'foo_bar', 'baz' ]` | `foo\_bar_baz` | `{"foo_bar":0,"baz":1}` |
| `[ 'foo.bar', 'baz' ]` | N/A | `{"foo":{"bar":0},"baz":1}` |

As you can see, instead of always being `true`, the value is now the offset of that field in the compound index.  This would be completely deterministic, it would allow us to perform index matching, and it's slightly more concise and easier to read.

[...]

PR #650, opened 2016-07-19
JSON-based index names

Tryneus:
Addresses #444, changes the index naming scheme to the following format:

`hz[_geo][_multi_<offset>]_<JSON>`

Where 'offset' is a number corresponding to the offset of the field that is multi-indexed.  We probably won't support having multiple multi fields in a single index as that would require generating all the permutations of multiple fields, which could be very expensive.  `geo` and `multi` aren't implemented yet as we do not have them in the API.  Still, the server should parse them fine and warn that they are not supported for when we add them later.

The JSON in the index name should be an array containing field names as strings, or nested fields as arrays of strings.

Example (from the comments in `index.js`):

```
Fields indexed: foo.bar, baz
Index name: hz_[["foo","bar"],"baz"]
```

This works around the problem with deterministic JSON serialization of objects mentioned in #444, and is slightly shorter.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/650)


marshall007:
:+1: good idea for the field name serialization.

Tryneus:
Merging after OTS review from @deontologician.

PR #653, opened 2016-07-20
handshake on multiplex subscription

### LINKED ISSUE #567: hz.currentUser doesn't execute until something else is executed

deontologician:
Fixes #567, which extended to all read queries: none of them would cause a handshake, so but they would block on the handshake, so they'd have to wait until a write query or an explicit `.subscribe` was called. Now we explicitly send the handshake on each request unless we've already sent it.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/653)

### DIFF
--- a/client/src/index.js
+++ b/client/src/index.js
@@ -5,7 +5,6 @@ import 'rxjs/add/operator/concatMap'
 import 'rxjs/add/operator/map'
 import 'rxjs/add/operator/filter'
 
-import * as Rx from 'rxjs'
 import { Collection, UserDataTerm } from './ast'
 import { HorizonSocket } from './socket'
 import { log, logError, enableLogging } from './logging'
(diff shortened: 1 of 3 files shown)


marshall007:
In that case, shouldn't we just always send the handshake lazily before the first query instead of also attempting it when the socket is opened?

marshall007:
... and then we could just make `hz.connect = () => socket.sendHandshake()`?

deontologician:
Good point, I'll make that change. Though we still have a potential problem when subscribing to the socket it may not actually send the handshake. There aren't a lot of good reasons to do that though, and we don't expose the socket externally. When we make the jump to typescript it'll definitely be private.

deontologician:
Ok, I made the fix and then reverted it. It doesn't properly emit the disconnected event when there's an error when you do this.

Issue #316, opened 2016-05-02
Remove usage of globals in client test suite


deontologician:
Right now we set all of the suite functions on global, but this is a leftover from when the tests only ran in the browser and didn't use preprocessing. Now the tests are preprocessed, and have access to the es6 module system, but don't make full use of it. We should:

1. change the `api.js` module to import the test suite modules explicitly
2. Modify the test suite modules so they no longer set a global variable

Unfortunately, there's still no way to eliminate globals that are injected by mocha (I believe), so those will have to remain.

PR #654, opened 2016-07-20
WIP: reorganize and refactor client tests

marshall007:
- Organize tests into sane folder structure.
- Tests should avoid setting and/or depending on global state.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/654)

### DIFF
--- a/client/test/query/time.js
+++ b/client/test/query/time.js
@@ -1,13 +1,13 @@
 import { _do as tap } from 'rxjs/operator/do'
 import { toArray } from 'rxjs/operator/toArray'
 
-import { assertCompletes, compareWithoutVersion } from './utils'
+import { assertCompletes, compareWithoutVersion } from '../utils'
 
-const timesSuite = global.timesSuite = getData => () => {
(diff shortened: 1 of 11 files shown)


client/test/query/time.js
```diff
@@ -1,13 +1,13 @@
...
+import { assertCompletes, compareWithoutVersion } from '../utils'
 
-const timesSuite = global.timesSuite = getData => () => {
+export default function() {
   let data
 
-  before(() => {
-    data = getData()
+  before(function() {
+    data = this.hz_data
```
marshall007, line 10: If we were not using arrow functions (which capture lexical scope) in all of our `it` assertions, we could reference `this.hz_data` and/or `this.horizon` in all our tests directly. This is unfortunate. Down the road it may be worth avoiding arrow functions in the tests so we can more easily manage their context and further avoid referencing things out of scope.

deontologician:
This will likely fix #316 

Looks awesome btw

PR #656, opened 2016-07-20
fix: properly change items with array ids

### LINKED ISSUE #649: Client doesn't refresh updated row properly ?
Using latest server/client w/ react.

1: I create new row via Horizon client @ store.

2: Update it via rethinkdb admin or python console, like:

`r.db("database").table("keywords_39c8a6fb745d").get(id).update({"randomField":55})`

3: Keep react app updated:

`horizon('keywords').watch().subscribe((kw) =>
    console.log('kw', kw));`

Line above returns 2 objects, old one(unupdated) and new one(it gives it id of -1). Is this desired behavior or bug?

It works if I refresh the page. Anyway to make it show just new object without having to refresh?

deontologician:
Fixes #649 

Added a unit test for it. Adds a dependency on `deep-equal`, and throws an error if we get a negative index for any reason, rather than silently corrupting the data

Thanks to @daniel1943 for patiently reproducing and @marshall007 for finding the bug

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/656)

### DIFF
--- a/client/src/ast.js
+++ b/client/src/ast.js
@@ -10,6 +10,7 @@ import 'rxjs/add/operator/take'
 import 'rxjs/add/operator/defaultIfEmpty'
 
 import snakeCase from 'snake-case'
+import deepEqual from 'deep-equal'
 
 import checkArgs from './util/check-args'
 import validIndexValue from './util/valid-index-value.js'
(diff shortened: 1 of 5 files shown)


client/src/ast.js
```diff
@@ -186,7 +187,14 @@ export function applyChange(arr, change) {
     } else {
       // If we don't have an offset, find the old val and
       // replace it with the new val
-      const index = arr.findIndex(x => x.id === change.old_val.id)
+      const index = arr.findIndex(x => deepEqual(x.id, change.old_val.id))
```
marshall007, line 190: @deontologician don't you need the same fix up in the `remove/uninitial` case on L161?

deontologician, line 190: good call, will open a new PR for this

deontologician, line 190: see #657

PR #658, opened 2016-07-20
serve.js script: On Windows, spawn `npm.cmd` instead of `npm`

danielmewes:
On Windows, there is no `npm` but only `npm.cmd`, at least when installing `node` using the official installer.
See http://stackoverflow.com/a/17537559/2413884 for details.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/658)

### DIFF
--- a/test/serve.js
+++ b/test/serve.js
@@ -78,8 +78,11 @@ const serve_file = (file_path, res) => {
   });
 };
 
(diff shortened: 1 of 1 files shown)


deontologician:
Does horizon otherwise work on windows with this fix?

danielmewes:
```
                                                                                  Yes
```

PR #669, opened 2016-07-22
Client test cleanups

deontologician:
- [X] remove function bind syntax
- [x] remove global usage and replace with imports ( #316 )
- [x] fix coverage so it's displayed over pre-babel source code, not the webpack output

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/669)

### DIFF
--- a/client/.gitignore
+++ b/client/.gitignore
@@ -13,6 +13,7 @@ lib-cov
 
 # Coverage directory used by tools like istanbul
 coverage
+.nyc_output
 
 # Grunt intermediate storage (http://gruntjs.com/creating-plugins#storing-task-files)
 .grunt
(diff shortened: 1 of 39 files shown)


deontologician:
Seems when you do `nyc report --reporter=html`, it seems to divide its time between reporting coverage for `lib/` (which we don't care about) and `/src` (which it reports only import statements as being covered for).  This is an annoying wart :/

deontologician:
Ok, remaining issue were from test.js importing `lib/index.js` for Horizon (the output of babel). So I aimed it at src/index.js. Also had to move around `.babelrc` and incorporate it into `package.json` for this to work. It looks like we have ~72% branch coverage, which isn't great but not as bad as it could have been

deontologician:
Circle ci broke as well :/

deontologician:
It works! We had a hardcoded reference to `src/.babelrc` in the webpack.horizon.js file which was causing a problem when it disappeared

PR #685, opened 2016-07-28
Reorganize horizon databases and tables

Tryneus:
This branch introduces a race condition when `auto_create_collection` is enabled by multiple horizon servers for the same project, but once ReQL supports it (rethinkdb/rethinkdb#5987), we can disambiguate name-conflicted tables.

There may still be some problems with the `schema` and `serve` commands, mostly because a lot of branches touched them recently, but I plan on testing them thoroughly in the near future.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/685)


server/src/metadata/collection.js
```diff
@@ -38,10 +29,23 @@ class Collection {
...
+      this.table = table.table;
+      this._table = table;
+      this._waiters.forEach((done) => this._table.on_ready(done));
+      this._waiters = [ ];
+    } else {
+      const old_table = this._table;
+      this.table = null;
+      this._table = null;
+
+      if (old_table) {
```
mlucy: It seems like if we need to do cleanup when removing a table here, we need to do it in the branch above as well.

server/src/metadata/metadata.js
```diff
@@ -353,14 +355,14 @@ class Metadata {
     error.check(this._collections.get(name) === undefined,
                 `Collection "${name}" already exists.`);
 
-    const collection = new Collection({ id: name, table: null }, this._db);
+    const collection = new Collection(name, null);
```
mlucy, line 360: Consider a comment.

server/src/schema/server_options.js
```diff
@@ -3,7 +3,7 @@
 const Joi = require('joi');
 
 const server = Joi.object({
-  project_name: Joi.string().default('horizon'),
+  project_name: Joi.string().default('project'),
```
mlucy: Consider changing this back to horizon.

mlucy:
LGTM

Tryneus:
Comments have been addressed, merging.

PR #690, opened 2016-07-28
CLI parsing rework

encryptio:
This changes the command module interface to push a bit more of the work to the modules themselves. It should make implementing plugins and the `cloud` subcommand simpler.

The new interface has one method, `runCommand`, which now takes the argument array after the command (instead of an `argparse.Namespace`), and `helpText` as it was before as a description of the command. `processConfig` and `addArguments` are no longer part of the command module API.

Some modules still have `processConfig`, but it's a function entirely local to the command module and is not called by `main.js`.

Pinging @Tryneus for review.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/690)

### DIFF
--- a/cli/src/main.js
+++ b/cli/src/main.js
@@ -6,6 +6,8 @@ process.title = 'horizon';
 
 const argparse = require('argparse');
 const chalk = require('chalk');
+const path = require('path');
+
 const initCommand = require('./init');
 const serveCommand = require('./serve');
 const versionCommand = require('./version');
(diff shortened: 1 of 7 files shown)


cli/src/main.js
```diff
@@ -29,42 +28,48 @@ const commands = {
...
-  };
-  try {
-    command.runCommand(options, done);
-  } catch (e) {
-    done(e);
-  }
+const cmdName = allArgs[0];
+const cmdArgs = allArgs.slice(1);
+
+if (cmdName == "-h" || cmdName == "--help" || cmdName == "help") {
```
dalanmiller, line 50: 👍 👍 👍

cli/src/version.js
```diff
@@ -8,13 +8,7 @@ const runCommand = () => {
   console.info(package_json.version);
```
Tryneus, line 8: if `create-cert` enforces that there are no extra arguments, shouldn't `version` as well?

cli/src/main.js
```diff
@@ -29,42 +28,48 @@ const commands = {
   schema: schemaCommand,
 };
 
-function parseArgs() {
-  const parser = new argparse.ArgumentParser();
-  const subparsers = parser.addSubparsers({
-    title: 'commands',
-    dest: 'command_name',
-  });
+const programName = path.basename(process.argv[1]);
```
Tryneus, line 31: Is `argv[1]` correct?  I imagine this will be different if you run it via `node <script> <subcommand>` vs `hz <subcommand>`.

Tryneus, line 31: Tbh, node probably does this right, nevermind.

Tryneus:
Looks good to me, merging.

PR #693, opened 2016-07-29
Added support for Horizon deploy to Google App Engine flexible environment

abelaska:
The Horizon javascript client optionally fetches Google Cloud Engine (GAE
flexible environment) external and internal IP address, completes the WS url
and only then connects to Horizon WS service.

This all happens only if the new option gaeInstanceIp is set with value 'external'
for GCE instance public or 'internal' for private IP address.

Horizon Google App Engine standard environment deployment is not supported.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/693)

### DIFF
--- a/client/src/index.js
+++ b/client/src/index.js
@@ -4,6 +4,7 @@ import 'rxjs/add/operator/catch'
 import 'rxjs/add/operator/concatMap'
 import 'rxjs/add/operator/map'
 import 'rxjs/add/operator/filter'
+import { Observable } from 'rxjs/Observable'
 
 import { Collection, UserDataTerm } from './ast'
 import { HorizonSocket } from './socket'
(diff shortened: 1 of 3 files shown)


client/src/index.js
```diff
@@ -22,15 +23,25 @@ function Horizon({
   lazyWrites = false,
   authType = 'unauthenticated',
   keepalive = 60,
+  gaeInstanceIp = undefined, // undefined, 'external', 'internal'
   WebSocketCtor = WebSocket,
 } = {}) {
   // If we're in a redirection from OAuth, store the auth token for
   // this user in localStorage.
 
+  let wsHost = host
```
deontologician, line 32: I think this logic should move out of the client, since it's specific to a particular vendor. It's easy enough to add this to a wrapper library

deontologician:
Hey @abelaska 

This is pretty cool, but I think we should keep any provider specific stuff out of the Horizon core libraries themselves. This would make an excellent plugin though.

I'm going to close this for now, but I've opened issue #694 to track making a plugin for this

abelaska:
Sure @deontologician no problem and thanks.

When will the plugin system be ready? Any ETA?

deontologician:
It is literally the next thing we're doing after the 2.0 release next week

PR #696, opened 2016-07-30
Fix CLI tests

Tryneus:
This is a bit of a refactor of the CLI because literally everything was broken by all the branches that just managed to land at about the same time.  The primary goal is to get the tests working, but I'm also addressing many inconsistencies across files in the cli module.

It's not quite done yet, but most of the hard work should be there.  As it stands, the scripts and tests were doing some pretty horrifying things, like returning early (and thus doing stuff concurrently), leaving rethinkdb running or connections open, using old functions that no longer exist, etc, etc.

Anyway, it's 7 AM, so I'll resume this tomorrow, but putting it up now in case anyone wants to make comments earlier on.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/696)

### DIFF
--- a/cli/src/serve.js
+++ b/cli/src/serve.js
@@ -1,7 +1,7 @@
 'use strict';
-const serve = module.exports = {};
 
 const chalk = require('chalk');
+const crypto = require('crypto');
 const fs = require('fs');
 const get_type = require('mime-types').contentType;
 const http = require('http');
@@ -9,14 +9,12 @@ const https = require('https');
 const open = require('open');
 const path = require('path');
 const argparse = require('argparse');
-const toml = require('toml');
 const url = require('url');
 
-const parse_yes_no_option = require('./utils/parse_yes_no_option');
+const config = require('./utils/config');
 const start_rdb_server = require('./utils/start_rdb_server');
-const interrupt = require('./utils/interrupt');
-const exit_with_error = require('./utils/exit_with_error');
-const is_directory = require('./utils/is_directory');
+const change_to_project_dir = require('./utils/change_to_project_dir');
+const shutdown = require('./utils/shutdown');
 const schema = require('./schema');
 
 const horizon_server = require('@horizon/server');
@@ -24,16 +22,12 @@ const logger = horizon_server.logger;
 
 const TIMEOUT_30_SECONDS = 30 * 1000;
 
-const default_config_file = '.hz/config.toml';
-const default_secrets_file = '.hz/secrets.toml';
 const default_rdb_host = 'localhost';
 const default_rdb_port = 28015;
 const default_rdb_timeout = 20;
 
-const helpText = 'Serve a Horizon app';
-
-function parseArguments(args) {
-  const parser = new argparse.ArgumentParser({prog: "hz serve"});
+const parseArguments = (args) => {
+  const parser = new argparse.ArgumentParser({ prog: 'hz serve' });
 
   parser.addArgument([ 'project_path' ],
     { type: 'string', nargs: '?',
@@ -78,6 +72,7 @@ function parseArguments(args) {
 
   parser.addArgument([ '--token-secret' ],
     { type: 'string', metavar: 'SECRET',
+      defaultValue: crypto.randomBytes(64).toString('base64'),
       help: 'Key for signing jwts. Default is random on each run' });
 
   parser.addArgument([ '--allow-unauthenticated' ],
@@ -132,7 +127,7 @@ function parseArguments(args) {
 
   parser.addArgument([ '--config' ],
     { type: 'string', metavar: 'PATH',
-      help: 'Path to the config file to use, defaults to "${default_config_file}".' });
+      help: 'Path to the config file to use, defaults to "${config.default_config_file}".' });
 
   parser.addArgument([ '--schema-file' ],
     { type: 'string', metavar: 'SCHEMA_FILE_PATH',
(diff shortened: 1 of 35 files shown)


Tryneus:
Oh, the biggest problem I noticed was that there was a circular dependency between `schema.js` and `serve.js`, which meant that the `serve` object was empty inside `schema` - meaning nothing should have ever worked - the tests were just so broken it didn't show up.

cli/test/schema.spec.js
```diff
@@ -31,90 +33,83 @@ indexes = ["datetime"]
...
-      });
-    });
-  });
+  let rdb_server;
+  let rdb_conn;
+
+  before('start rethinkdb', () =>
+    start_rdb_server({
+      bind: ["127.0.0.1"],
+      dataDir: tmpdir() + "/horizon-test",
```
dalanmiller: One thing I forgot to add here was we should save this generated path and then remove it after the tests are done or have exited. 

Maybe this?:

``` js
describe("hz schema", () => {
  let rdb_server, rdb_conn;
  const tmpPath = tmpdir() + "/horizon-test"; 

  before('start rethinkdb', () => 
    start_rdb_server({
      bind: ["127.0.0.1"],
      dataDir: tmpPath,
    }).then((server) => {
      rdb_server = server;
    });
  );

  after('stop rethinkdb', () => {
    // Delete temporary database dir
    fs.rmdirSync(tmpPath);
    // Close RDB server connection
    return rdb_server.close();
  });

  ...
```

cli/test/schema.spec.js
```diff
@@ -31,90 +33,83 @@ indexes = ["datetime"]
...
-    it("saves schema to schema.toml from rdb", () => {
-      return runSaveCommand({
+    it("saves schema to schema.toml from rdb", () =>
+      runSaveCommand({
         start_rethinkdb: false,
         rdb_host: '127.0.0.1',
         rdb_port: currentOpts.dbPort,
-        out_file: fs.createWriteStream('out.toml', {flags: 'w'}),
+        out_file: fs.createWriteStream('out.toml', { flags: 'w' }),
         project_name: 'horizon_schema_test',
```
dalanmiller: This line should just be `project_name` using the `const` defined at the top

cli/test/schema.spec.js
```diff
@@ -125,66 +120,39 @@ describe("hz schema", () => {
...
```
dalanmiller: This should also probably use the `project_name` `const`.

dalanmiller:
@Tryneus you must be Midas, because all these tests have been turned to gold.

cli/test/schema.spec.js
```diff
@@ -31,90 +33,83 @@ indexes = ["datetime"]
...
-    
-    beforeEach('set up database state', () => {
-      // Create MockFS
-      mockFs({
-          '.hz': {
-            'schema.toml': testSchema,
-          },
-          'out.toml': '',
-      });
+    //beforeEach('initialize mockfs', () => {
```
dalanmiller: Should this be commented out?

Tryneus:
All the tests should be more or less passing at this point.  Some of the cli tests can be rather slow as they involve creating several RethinkDB tables in the setup phase.

cli/test/schema.spec.js
Tryneus: +1, fixed in the latest upload

cli/test/schema.spec.js
Tryneus: fixed in the latest upload

cli/test/schema.spec.js
Tryneus: I added a stage to clean up the directory after the suite runs.

cli/test/schema.spec.js
Tryneus: Ah, I wasn't completely sure if I'd be breaking anything by removing this as I'd added a mockFs initialization higher up in the suite.  Turns out this was no longer necessary, so I removed it.

cli/src/utils/initialize_joi.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+// Issues a dummy joi validation to force joi to initialize its scripts.
+// This is used because tests will mock the filesystem, and the lazy
+// `require`s done by joi will no longer work at that point.
+module.exports = (joi) =>
+  joi.validate('', joi.any().when('', { is: '', then: joi.any() }));
```
dalanmiller, line 7: According to @marsup, calling `Joi.options()` would also be sufficient.

https://gitter.im/hapijs/joi?at=57a1236747659bfb108977f7

Marsup, line 7: Just checked, `Joi.options({})` actually.

dalanmiller:
A nitpick: it seems that I can no longer use `console.log` in the test files to output some debugging variables as the logging is being squelched somewhere. 

I also couldn't even do `process.stdout.write('something')` and have it appear in the output.

cli/src/utils/initialize_joi.js
dalanmiller, line 7: Thanks @Marsup! 👍

cli/src/serve.js
```diff
@@ -132,7 +127,7 @@ function parseArguments(args) {
 
   parser.addArgument([ '--config' ],
     { type: 'string', metavar: 'PATH',
-      help: 'Path to the config file to use, defaults to "${default_config_file}".' });
+      help: 'Path to the config file to use, defaults to "${config.default_config_file}".' });
 
   parser.addArgument([ '--schema-file' ],
     { type: 'string', metavar: 'SCHEMA_FILE_PATH',
```
dalanmiller, line 133: Extra `"` in help for `--schema-file`

deontologician:
:lgtm: mergin'

Issue #712, opened 2016-08-03
In the next branch if server errors and closes connection on handshake client error is not called


nphyatt:
I was looking into making a PR for #709 and it seemed the easiest course of action would be to close the connection with an error on the handshake if the connection limit had been reached. However I noticed if the handshake responds with an error and closes the connection the client observables error function doesn't actually get called anymore. I'm guessing it's coming from the refactor on [this commit](https://github.com/rethinkdb/horizon/commit/46dfcb075a182408677560723a7902937c1cdfc7) 

No matter what response the `sendHandshake()` method gets from the server [here](https://github.com/rethinkdb/horizon/blob/next/client/src/socket.js#L133) the `next()` function is always called and not the error.

It does seem the `hzRequest()` method adds some checking [here](https://github.com/rethinkdb/horizon/blob/next/client/src/socket.js#L178) to see if a response contains an error but only on messages sent after the handshake...perhaps this could be added to the handshake response as well?

This can be verified by connecting with an invalid JWT. The error function [here](https://github.com/rethinkdb/horizon/blob/next/client/src/index.js#L49) never gets called.

I started diving into it but it's late and I'm tired. I'll look into it more tomorrow.

danielmewes:
Sounds like a bug. Thanks for reporting this and for the PR.

PR #713, opened 2016-08-03
Handle Errors in deserialize function

nphyatt:
Fix for Issue #712 so that any message with an error key in the top level of the message throws an error in the client.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/713)

### DIFF
--- a/client/src/serialization.js
+++ b/client/src/serialization.js
@@ -3,12 +3,22 @@ const PRIMITIVES = [
 
 function modifyObject(doc) {
   Object.keys(doc).forEach(key => {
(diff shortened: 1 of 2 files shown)


danielmewes:
Thanks for the pull request @nphyatt !

@deontologician

deontologician:
This causes a bunch of test failures. I think this is from the protocol error being thrown in the deserializer. Since we deserialize all messages on the socket, errors in the deserializer don't propagate to the observables for each request (they die before being routed there). So I think this will need a little more work to fix. I'll check it out today

Issue #722, opened 2016-08-04
Refactor cli arg parsing


deontologician:
Right now there's lots of duplicated copy/pasted code, and it's hard to follow. The ideal situation would be specifying in one place for each option:
### Option specification
1. The argument name
2. All argparse options
3. the default value
4. A namespace if necessary (`--rdb-*`, or `--auth-*`)
5. a function for any special post-processing
   - `bind=all` turns into `bind: [ '0.0.0.0' ]`
   - `connect` needs to be parsed afterward
   - if this post-processor fails, processing should stop and the exception should be shown in red
6. Whether the option is considered a secret (and should come from `secrets.toml`) (default:no)
7. Whether the option should be part of --dev (default: no)
   - this means every command can potentially have a `--dev` option, but they don't need to use it
8. Whether the option should be specifiable in the config (default: yes)
9. Whether the option should be specifiable in the command line (default: yes)
10. Whether the option should be specifiable as an env variable
11. An optional validation group for this option

Additionally, a validation function will need to be able to be associated with the validation group so that commands that potentially conflict (like --rdb-port and --connect) can be validated together once everything is merged into a final options object. The function would receive all values of commands in that group, even ones that defaulted.
### Features:
1. it should handle our `yes|no` boolean stuff transparently. If you specify an option `type: 'bool'` it will accept yes/no, there shouldn't be a need to do any more special checks.
2. it should handle env/config/secrets/option merging completely transparently, that can get out of the individual commands
3. it should automatically add any dev options to the `--dev` option help text and handle the logic for adding dev options
4. It should be easy to pull out reuseable options like rethinkdb options and auth options and just extend the per-command options very simply. This will also help plugin writers.
### misc cleanups
- `_implicit_start_rethinkdb` and `_dev_flag_used` express the same information and should be merged.

PR #728, opened 2016-08-06
feature: hz migrate

### LINKED ISSUE #705: `hz migrate` for 2.0
Add a new command `hz migrate` to the CLI, that connects to an existing RethinkDB server (or starts one, depending on command line arguments), and then migrates the metadata in RethinkDB to the new 2.0 format (renaming indexes, deleting the internal DB at the end etc.).

When it's done, it should tell the user that it migrated the schema and suggest running `hz schema save` to save the new schema. Or alternatively, it can run `hz schema save` itself and print where it has saved the schema.
### Todos

(added by @deontologician):
- [x] Make a backup of the database
  - [x] Add an option to turn this off
- [x] Move tables from `${project}_internal.foo` to `${project}.hz_foo`
  - [x] Don't add the `hz_` prefix to the `users` table
- [x] Rename all indices from the previous format to the json format
- [x] Fix the `hz_table` format from `{ id: collectionName,  table: tableName }` to `{ id: collectionName, table_id: tableId }`
- [x] Rename tables to match the collection name
- [x] Export the schema
  - [ ] Add an option to turn this off

deontologician:
Migrates from pre 2.0 to 2.0 format only
Fixes #705

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/728)

### DIFF
--- a/cli/src/migrate.js
+++ b/cli/src/migrate.js
@@ -1,0 +1,373 @@
+'use strict';
+const chalk = require('chalk');
+const r = require('rethinkdb');
+const Promise = require('bluebird');
+const argparse = require('argparse');
+const child_process = require('child_process');
+const runSaveCommand = require('./schema').runSaveCommand;
+const fs = require('fs');
+const accessAsync = Promise.promisify(fs.access);
+const config = require('./utils/config');
+const interrupt = require('./utils/interrupt');
+const change_to_project_dir = require('./utils/change_to_project_dir');
+const parse_yes_no_option = require('./utils/parse_yes_no_option');
+const start_rdb_server = require('./utils/start_rdb_server');
+
+const VERSION_2_0 = [ 2, 0, 0 ];
+
+function run(cmdArgs) {
+  const options = processConfig(cmdArgs);
+  interrupt.on_interrupt(() => teardown());
+  return Promise.resolve().bind({ options })
+    .then(setup)
+    .then(validateMigration)
+    .then(makeBackup)
+    .then(renameUserTables)
+    .then(moveInternalTables)
+    .then(renameIndices)
+    .then(rewriteHzCollectionDocs)
+    .then(exportNewSchema)
+    .finally(teardown);
+}
+
+function green() {
+  const args = Array.from(arguments);
+  args[0] = chalk.green(args[0]);
+  console.log.apply(console, args);
+}
+
+function white() {
+  const args = Array.from(arguments);
+  args[0] = chalk.white(args[0]);
+  console.log.apply(console, args);
+}
+
+function processConfig(cmdArgs) {
+  // do boilerplate to get config args :/
+  const parser = new argparse.ArgumentParser({ prog: 'hz migrate' });
+
+  parser.addArgument([ 'project_path' ], {
+    default: '.',
+    nargs: '?',
+    help: 'Change to this directory before migrating',
+  });
+
+  parser.addArgument([ '--project-name', '-n' ], {
+    help: 'Name of the Horizon project server',
+  });
+
+  parser.addArgument([ '--connect', '-c' ], {
+    metavar: 'host:port',
+    default: undefined,
+    help: 'Host and port of the RethinkDB server to connect to.',
+  });
+
+  parser.addArgument([ '--rdb-user' ], {
+    default: 'admin',
+    metavar: 'USER',
+    help: 'RethinkDB User',
+  });
+
+  parser.addArgument([ '--rdb-password' ], {
+    default: undefined,
+    metavar: 'PASSWORD',
+    help: 'RethinkDB Password',
+  });
+
+  parser.addArgument([ '--start-rethinkdb' ], {
+    metavar: 'yes|no',
+    default: 'yes',
+    constant: 'yes',
+    nargs: '?',
+    help: 'Start up a RethinkDB server in the current directory'
+  });
+
+  parser.addArgument([ '--config' ], {
+    default: '.hz/config.toml',
+    help: 'Path to the config file to use, defaults to ".hz/config.toml".',
+  });
+
+  parser.addArgument([ '--skip-backup' ], {
+    metavar: 'yes|no',
+    default: 'no',
+    constant: 'yes',
+    nargs: '?',
+    help: 'Whether to perform a backup of rethinkdb_data before migrating',
+  });
+
+  const parsed = parser.parseArgs(cmdArgs);
+  const confOptions = config.read_from_config_file(parsed.project_path, parsed.config);
+  const envOptions = config.read_from_env();
+  config.merge_options(confOptions, envOptions);
+  // Pull out the relevant settings from the config file
+  const options = {
+    project_path: parsed.project_path || '.',
+    project_name: parsed.project_name || confOptions.project_name,
+    rdb_host: parsed.rdb_host || confOptions.rdb_host || 'localhost',
+    rdb_port: parsed.rdb_port || confOptions.rdb_port || 28015,
+    rdb_user: parsed.rdb_user || confOptions.rdb_user || 'admin',
+    rdb_password: parsed.rdb_password || confOptions.rdb_password || '',
+    skip_backup: parse_yes_no_option(parsed.skip_backup),
+    start_rethinkdb: parse_yes_no_option(parsed.start_rethinkdb),
+  };
+  // sets rdb_host and rdb_port from connect if necessary
+  if (parsed.connect) {
+    config.parse_connect(parsed.connect, options);
+  }
+
+  if (options.project_name == null) {
+    throw new Error('No project_name given');
+  }
+  return options;
+}
+
+function setup() {
+  // Start rethinkdb server if necessary
+  // Connect to whatever rethinkdb server we're using
+  white('Setup');
+  return Promise.resolve().then(() => {
+    if (this.options.project_path && this.options.project_path !== '.') {
+      green(` ├── Changing to directory ${this.options.project_path}`);
+      change_to_project_dir(this.options.project_path);
+    }
+  }).then(() => {
+    // start rethinkdb server if necessary
+    if (this.options.start_rethinkdb) {
+      green(' ├── Starting RethinkDB server');
+      return start_rdb_server({ quiet: true }).then((server) => {
+        this.rdb_server = server;
+        this.options.rdb_host = 'localhost';
+        this.options.rdb_port = server.driver_port;
+      });
+    }
+  }).then(() => {
+    green(' ├── Connecting to RethinkDB');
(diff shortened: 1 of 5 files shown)


deontologician:
ping @Tryneus for review

cli/src/migrate.js
```diff
@@ -0,0 +1,354 @@
...
+    default: '.hz/config.toml',
+    help: 'Path to the config file to use, defaults to ".hz/config.toml".',
+  });
+
+  parser.addArgument([ '--skip-backup' ], {
+    action: 'storeTrue',
+    help: 'Whether to perform a backup of rethinkdb_data before migrating',
+  });
+
+  const parsed = parser.parseArgs(cmdArgs);
```
Tryneus, line 98: This should be doing similar things to the other scripts - that is, read from the config files, read from env, read from command-line, and merge them together in that order.  Currently, this doesn't read from env and implements precedence on an option-by-option basis, which seems bad.

cli/src/migrate.js
```diff
@@ -0,0 +1,354 @@
...
+    .then(makeBackup)
+    .then(renameUserTables)
+    .then(moveInternalTables)
+    .then(renameIndices)
+    .then(rewriteHzCollectionDocs)
+    .then(exportNewSchema)
+    .then(teardown)
+    .catch((err) => {
+      red(` └── ${err.message}`);
+      return Promise.bind(options).then(teardown());
```
Tryneus: `main.js` already reports errors and exits with a non-zero exit code (which this won't do)

cli/src/migrate.js
```diff
@@ -0,0 +1,354 @@
...
+    skip_backup: parsed.skip_backup || false,
+    start_rethinkdb: parsed.start_rethinkdb || false,
+  };
+  // sets rdb_host and rdb_port from connect if necessary
+  if (parsed.connect) {
+    config.parse_connect(parsed.connect, options);
+  }
+
+  if (options.project_name == null) {
+    throw new Error('project_name is null');
```
Tryneus: This error message seems a little misleading - I imagine this indicates that `project_name` has not been set, may as well say that.

cli/src/migrate.js
```diff
@@ -0,0 +1,354 @@
...
+
+function teardown() {
+  return Promise.resolve().then(() => {
+    white('Cleaning up...');
+    // close the rethinkdb connection
+    if (this.conn) {
+      green(' ├── Closing rethinkdb connection');
+      return this.conn.close();
+    } else {
+      return undefined;
```
Tryneus: no need to return undefined

Tryneus: line 168 as well

Tryneus: and 136

cli/src/migrate.js
```diff
@@ -0,0 +1,354 @@
...
+    action: 'storeTrue',
+    help: 'Start up a RethinkDB server in the current directory'
+  });
+
+  parser.addArgument([ '--config' ], {
+    default: '.hz/config.toml',
+    help: 'Path to the config file to use, defaults to ".hz/config.toml".',
+  });
+
+  parser.addArgument([ '--skip-backup' ], {
```
Tryneus, line 90: `--start-rethinkdb` and `--skip-backup` should probably be yes/no options so that they can be overridden in either direction.  Other scripts should have it configured this way already (for `--start-rethinkdb`).

Tryneus, line 90: Note that this doesn't mean that users _have_ to specify yes or no, you can still have the option not require a value.

cli/src/migrate.js
```diff
@@ -0,0 +1,354 @@
...
+          .branch(r.error(`Some tables in ${project} have an hz_ prefix`), true)
+
+  return Promise.resolve().then(() => {
+    white('Validating current schema version');
+    return r.dbList().contains(`${project}_internal`)
+      .branch(true, r.error(`${project}_internal not found`))
+      .do(() => checkForHzTables)
+      .run(this.conn)
+      .then(() => green(' └── Pre-2.0 schema found'))
+      .catch(() => {
```
Tryneus: This could fail for a number of reasons, should probably wrap the error rather than discard it.

cli/src/migrate.js
```diff
@@ -0,0 +1,354 @@
...
+        proc.stderr.setEncoding('utf8');
+        const err = proc.stderr.read();
+        reject(new Error(`rethinkdb dump exited with an error:\n\n${err}`));
+      }
+    });
+  });
+}
+
+function renameUserTables() {
+  // for each table listed in ${project}_internal.tables
```
Tryneus: it's `${project}_internal.collection`, not `${project}_internal.tables`

cli/src/migrate.js
```diff
@@ -0,0 +1,354 @@
...
+    white('Renaming indices to new JSON format');
+    return r.db(project).tableList().forEach((tableName) => {
+      return r.db(project).table(tableName).indexList().forEach((indexName) => {
+        return r.db(project).table(tableName).indexRename(indexName, rename(indexName));
+      });
+    }).run(this.conn)
+    .then(() => green(' └── Indices renamed.'));
+  });
+
+  function rename(name) {
```
Tryneus, line 294: It wasn't _necessary_ to do this in ReQL.  Less risky to just reuse the old functions.

Tryneus, line 294: Still, pretty interesting

cli/src/migrate.js
```diff
@@ -0,0 +1,354 @@
...
+      })
+      // last field needs to be appended
+      .do((x) => x('fields').append(x('field')))
+      .toJSON()
+      .do((x) => r('hz_').add(x));
+  }
+}
+
+function rewriteHzCollectionDocs() {
+  // for each document in ${project}.hz_tables
```
Tryneus: `hz_collections`, not `hz_tables`

cli/src/migrate.js
```diff
@@ -0,0 +1,354 @@
...
+  }).then(() => {
+    green(' ├── Connecting to RethinkDB');
+    return r.connect({
+      host: this.options.rdb_host,
+      port: this.options.rdb_port,
+      user: this.options.rdb_user,
+      password: this.options.rdb_password,
+    });
+  }).then((conn) => {
+    green(' └── Successfully connected');
```
Tryneus, line 152: Should probably add a step to wait for the tables to be ready.  I noticed when you were testing you used an external rethinkdb server, which usually doesn't have this problem.

Tryneus, line 152: Specifically, we need `hz_collections` to be ready, I don't think we touch the data in any of the others.

cli/src/migrate.js
```diff
@@ -0,0 +1,354 @@
...
+  // Import and run schema save process, giving it a different
+  // filename than schema.toml
+  return Promise.resolve().then(() => {
+    white('Exporting the new schema to .hz/schema.toml.migrated');
+    return runSaveCommand({
+      rdb_host: this.options.rdb_host,
+      rdb_port: this.options.rdb_port,
+      rdb_user: this.options.rdb_user,
+      rdb_password: this.options.rdb_password,
+      rdb_timeout: this.options.rdb_timeout,
```
Tryneus: `rdb_timeout` isn't in the listed options above

cli/src/migrate.js
deontologician, line 98: I felt like env vars would be less useful for this. I think this implements precedence the same way as the others, config file first, then command line afterward.

I only grab a few specific items from the config file, because in general the config is for `hz serve`, we don't have any way of calling out "this option should be used when migrating only". Same thing goes for env vars.

cli/src/migrate.js
deontologician: It wasn't doing cleanup for whatever reason, but I guess I can use `finally` for that

cli/src/migrate.js
deontologician: This was just to appease the linter. It doesn't like when you implicitly return undefined. In hindsight, this is stupidly verbose, I'll tweak the eslint settings.

cli/src/migrate.js
deontologician, line 90: I looked at the yes/no handling and it seemed annoying so I made them normal flag variables. I can set this up though since I have more time today.

cli/src/migrate.js
deontologician: :+1:

cli/src/migrate.js
deontologician, line 294: Totally necessary!

cli/src/migrate.js
deontologician: Whoops, these line notes were from the original sketch, I didn't update them after I wrote the code

cli/src/migrate.js
Tryneus, line 98: I don't think we can just ignore environment variables like that.  If a user's setup involves setting RethinkDB options in the env (and not by config file or command-line), then this will break all their assumptions.

Tryneus, line 98: Also, the `config` stuff is supposed to be generic enough to work for any of the scripts, not just `serve`.  It may not be completely there yet, but you should be able to use it without any modifications.

deontologician, line 98: I believe it, but I don't want to mess with it too much since it's going to change with #722

cli/src/migrate.js
deontologician: ok, it wasn't eslint, it was js2-mode. I turned off the error!

cli/src/migrate.js
```diff
@@ -0,0 +1,351 @@
...
+    .then(makeBackup)
+    .then(renameUserTables)
+    .then(moveInternalTables)
+    .then(renameIndices)
+    .then(rewriteHzCollectionDocs)
+    .then(exportNewSchema)
+    .finally(teardown);
+}
+
+function green(arg, ...args) {
```
Tryneus: I don't think this syntax works in all `node` versions we're supporting.  You may need to use `apply` with `arguments`.

cli/src/migrate.js
deontologician, line 98: Ok, I give up, I will parse env options

cli/src/migrate.js
```diff
@@ -170,23 +170,37 @@ function teardown() {
 function validateMigration() {
   // check that `${project}_internal` exists
   const project = this.options.project_name;
+  const niceErr = (e) => {
```
Tryneus: pretty sure you can just use `err.msg` if it's a reql error if you want to exclude the stack trace.

Tryneus:
Looks good, merge at will.

deontologician:
Tests they are a-passin'. PR I am a-mergin'.

Issue #727, opened 2016-08-05
Move rdb_password to secrets.toml


deontologician:
It's a secret

PR #731, opened 2016-08-09
Connection options in template toml files

Tryneus:
Addresses #727.  The toml file parsing doesn't actually distinguish between `config.toml` and `secrets.toml`, so the options can really go in either.  Added some comments to the templates in `init.js` to encourage safe use (and added missing options).

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/731)


deontologician:
:+1:

Issue #706, opened 2016-08-02
New index specification format for `hz schema`


danielmewes:
Currently a schema `.toml` file contains index specifications as strings. We decided to start using a structured index specification in Horizon 2.0, to make these easier and more explicit to write and read.

@Tryneus can you post some details on the new format please for reference once you have a full spec?

Tryneus:
The format should look like this, for an example collection 'foo' with two indexes, one on the nested field 'a.b' and regular field 'c', and the other on the fields 'bar' and 'baz':

```
[collections.foo]
[[collections.foo.indexes]]
fields = [['a','b'],['c']]
[[collections.foo.indexes]]
fields = [['bar'],['baz']]
```

Working on a PR now.

PR #732, opened 2016-08-09
Modify schema index format

Tryneus:
Addresses #706.  Also fixes a problem with non-nested fields.  Specifically, TOML cannot generate an array with disparate types (i.e. strings for fields and arrays for nested fields for a given index).  Therefore, the index fields format is now unified so that is it always an array of arrays.  This is probably better as it ensures there is exactly one format for an index name (barring whitespace concerns).

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/732)

### DIFF
--- a/cli/src/schema.js
+++ b/cli/src/schema.js
@@ -21,7 +21,6 @@ const toml = require('toml');
 const r = horizon_server.r;
 const create_collection_reql = horizon_metadata.create_collection_reql;
 const initialize_metadata = horizon_metadata.initialize_metadata;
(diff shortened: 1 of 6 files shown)


deontologician:
A bunch of tests are failing related to serialized dates ( plus a few others)

danielmewes:
@chipotle The documentation will need another update:

> TOML cannot generate an array with disparate types (i.e. strings for fields and arrays for nested fields for a given index). Therefore, the index fields format is now unified so that is it always an array of arrays.

PR #734, opened 2016-08-10
Add missing rxjs imports

KeKs0r:
I was trying to test 2.0 with my local project and realized that those requires are missing.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/734)


deontologician:
Excellent! Thanks, we'll merge this today

deontologician:
This would actually have been caught by typescript... hmm. Gotta get on that.

Anyways, thanks again @KeKs0r

PR #741, opened 2016-08-12
Definitely Typescript

deontologician:
This is going to allow us to add and package typescript, I'll probably move modules incrementally.

Typescript 2.0
Webpack 2.1+

Also had to bump babel so we could get it to do es2015 for everything except modules (since webpack 2.0 has tree shaking with es6 module syntax)
### TODO
- [x] port test webpack config over to webpack 2.0
- [x] fix npm scripts to use new webpack command line env options instead of cross-env
- [x] port all utils modules to typescript

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/741)


client/webpack.config.js
```diff
@@ -1,40 +1,153 @@
...
+function determinePlugins(options) {
+  const plugins = [
+    new BannerPlugin('__LICENSE__'),
   ]
+  if (options.devBuild) {
+    plugins.push(new DedupePlugin())
+    plugins.push(new OccurrenceOrderPlugin())
+    plugins.push(new UglifyJsPlugin({
+      compress: {
+        screw_ie8: false,
```
dalanmiller, line 180: Best option name ever.

dalanmiller:
Should we also define our custom Horizon types in this PR and also rip out Joi?

deontologician:
I dont think we have joi in the client now do we?

dalanmiller:
Ahh I assumed we did, cause it's everywhere else.

client/webpack.config.js
vramana, line 180: Totally agreed :100:

deontologician:
UNfortunately, coverage seems broke

deontologician:
Blargh, broke the tests again

niieani:
TypeScript 2.0 final was just released :)

PR #746, opened 2016-08-12
Set log level on server construction

KeKs0r:
This PR allows to change the log level for embedded horizon.

This is not yet tested. Wanted to get feedback first. If this approach is OK. I will run some additional tests.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/746)

### DIFF
--- a/cli/src/serve.js
+++ b/cli/src/serve.js
@@ -320,15 +320,14 @@ const start_horizon_server = (http_servers, opts) =>
     rdb_user: opts.rdb_user || null,
     rdb_password: opts.rdb_password || null,
     rdb_timeout: opts.rdb_timeout || null,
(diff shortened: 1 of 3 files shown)


deontologician:
ping @Tryneus . My understanding was that you can set winston log levels globally from the winston library. Is that incorrect? I think generally it's bad to pollute things with logging parameters since logging is such a cross-cutting concern. But I could be wrong here...

Tryneus:
@deontologician - yes, this doesn't have to be an option for the server constructor, the horizon server module exposes `logger` already.

KeKs0r:
if the logger is exported, it can be changed from the embedding script. Thanks for the clarification.

Issue #680, opened 2016-07-27
SSL option for RethinkDB connection


danielmewes:
https://github.com/rethinkdb/horizon/pull/507 added support for RethinkDB's user and password options, but didn't add a way for enabling SSL for the RethinkDB connection.

We'll need an option to pass the location of the RethinkDB SSL certificate to Horizon, and Horizon has to pass the certificate's file path to `r.connect` through the `ssl` option if specified.

This is required for connecting directly to Compose RethinkDB clusters for example.

geddski:
+1

Linicks:
+1 For me as well.  Unfortunately, this became a non-starter for me because my company has a secure protocol only policy.  Hopefully, this can be addressed soon, because I can see allot of promise in Horizon.

GeoffreyPlitt:
+1 I thought 2.0 was going to support SSL?

danielmewes:
We should put this into Horizon 2.1.0 (or maybe even still add it to 2.0.0 before we release the final version?)

GeoffreyPlitt:
Cool!

b-d:
This was a problem for me as well, it's a quick fix to add a `rdb_cert` option in the `horizon()` constructor and then add 
`ssl: {
                    ca: this._cert
          }`
to the `r.connect()` call in 
https://github.com/rethinkdb/horizon/blob/next/server/src/reql_connection.js

GeoffreyPlitt:
Is there an ETA yet on Horizon+Compose compatibility?

PR #758, opened 2016-08-17
feature: allow passing in rdb ssl cert to Server

deontologician:
Based on @b-d 's comment here: https://github.com/rethinkdb/horizon/issues/680#issuecomment-240522809

This just adds it to the server, not to `hz serve` options etc. We could add it if people want.

Also, I didn't test or write tests for this.

ping @Tryneus for review

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/758)

### DIFF
--- a/server/src/reql_connection.js
+++ b/server/src/reql_connection.js
@@ -12,7 +12,7 @@ class ReqlConnection {
   constructor(host, port, db,
               auto_create_collection, auto_create_index,
               user, pass, connect_timeout,
-              interruptor) {
+              interruptor, ca) {
     this._rdb_options = {
       host,
       port,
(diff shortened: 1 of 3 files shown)


deontologician:
The test failures indicate I should probably not allow providing null as an ssl cert

Tryneus:
Looks good, once you fix the test issue.

danielmewes:
Would it be hard to add the flags to the CLI modules?
It feels pretty incomplete to have it only when instantiating the server manually.

I guess we could add the CLI (and config-file) flags later for Horizon 2.1.0 though.

lirbank:
Hi guys, any chance you could prioritize getting this PR merged and make a new release? Not being able to connect to our Compose DB is a blocker so would be nice to have it resolved, even if in a temporary fashion. 

Btw, are there any other hosting options for RethinkDB than Compose, while waiting for Horizon Cloud?

danielmewes:
I re-triggered the test on CircleCI and it passed this time. @deontologician is there more to do here (apart from maybe adding the CLI option)?

PR #767, opened 2016-08-18
Fix group deletion behavior in server

Tryneus:
While porting code to the permissions plugin, I noticed this logic was wrong, would be nice to have it fixed until plugins hit.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/767)


deontologician:
:lgtm:

Issue #766, opened 2016-08-18
Make Horizon.hasAuthTokens() check whether the token is expired


deontologician:
Right now when a token expires, a page refresh is pretty much needed. We don't give any way of hooking into the event to detect a failure to authenticate, so the experience is really bad.

We should make `Horizon.hasAuthToken()` check whether the jwt is expired. Without including the server public key in our JWTs (which we don't) a user could spoof the expiration date in the jwt, but in the worst case it should just cause us to get an error from the server (which does validate the signature).

deontologician:
Fixed by #768

PR #768, opened 2016-08-19
Check if jwt token is expired

KeKs0r:
This is my attempt to solve #766

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/768)

### DIFF
--- a/client/src/auth.js
+++ b/client/src/auth.js
@@ -127,19 +127,18 @@ export class TokenStorage {
 
   // Whether there is an auth token for the provided authType
   hasAuthToken() {
-    const token = this.get();
-    if(!Boolean(token)){
-      return false;
+    const token = this.get()
+    if (!token) {
+      return false
     }
     try {
-      const meta = JSON.parse(atob(token.split('.')[1]));
-      const exp = meta.exp;
-      const now = new Date().getTime() / 1000;
-      return (now < exp);
-    } catch (e){
-      return false;
+      const meta = JSON.parse(atob(token.split('.')[1]))
+      const exp = meta.exp
+      const now = new Date().getTime() / 1000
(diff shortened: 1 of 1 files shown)


client/src/auth.js
```diff
@@ -127,7 +127,19 @@ export class TokenStorage {
...
+    }
+    try {
+      const meta = JSON.parse(atob(token.split('.')[1]));
+      const exp = meta.exp;
+      const now = new Date().getTime() / 1000;
+      return (now < exp);
+    } catch (e){
+      return false;
+    }
+    return Boolean(token)
```
deontologician: I don't think this line is reachable

client/src/auth.js
```diff
@@ -127,7 +127,19 @@ export class TokenStorage {
 
   // Whether there is an auth token for the provided authType
   hasAuthToken() {
-    return Boolean(this.get())
+    const token = this.get();
+    if(!Boolean(token)){
```
deontologician: `!` will coerce to a boolean, so you can just do `if (!token) {`

client/src/auth.js
```diff
@@ -127,7 +127,19 @@ export class TokenStorage {
...
+      return false;
+    }
+    try {
+      const meta = JSON.parse(atob(token.split('.')[1]));
```
deontologician: If you run eslint it will call out a few formatting errors like extraneous semicolons

deontologician:
Logic looks good @KeKs0r, just a few minor comments to address. Thanks for the PR!

KeKs0r:
@deontologician Have implemented your comments.

deontologician:
Awesome, thanks a lot :)

PR #770, opened 2016-08-20
Fix, setAuthFromQueryParams() compatible with React Native (android)

lirbank:
`window` exists on Android but does not have a `location` property, so
it’s necessary to check that `window.location` is not null before trying to access `window.location.search`.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/770)


deontologician:
Thanks, looks good :+1:

PR #776, opened 2016-08-23
Cli error cleanups

### LINKED ISSUE #761: We should give a better error if you run `hz serve` without a PEM file.
Currently we produce this output:

```
serve failed with Error: Could not access file "./horizon-key.pem" for running HTTPS server: Error: ENOENT: no such file or directory, open '/home/mlucy/hzdemo/horizon-key.pem'
```

It would be nice if we gave an error that:
- Explains more clearly what went wrong if you don't know what a PEM file is.
- Recommends running with `--dev` during development.
- Possibly tells you how generate a self-signed cert for testing?
- Possibly links to a page explaining what certificates are and how to generate a self-signed one for testing?

deontologician:
Fixes #761 

As a part of it, I added a new `NiceError` class that formats errors better, and modified the main cli catch clause to use it.

Also caught a bug where some command flag options were being ignored.

The original goal of #761 was to give a better error when doing `hz serve` instead of `hz serve --dev`, since the exception you'll get is something about `horizon-cert.pem` and is confusing if you just want to get started. Now, if you do `hz serve`, you'll get:

selection_090

If you actually called `hz serve --cert-file=some_filename.pem`, it gives a more terse error:

selection_089

Assigning to @mlucy since he was the one who originally complained :P

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/776)

### DIFF
--- a/cli/src/serve.js
+++ b/cli/src/serve.js
@@ -14,6 +14,7 @@ const url = require('url');
 const config = require('./utils/config');
 const start_rdb_server = require('./utils/start_rdb_server');
 const change_to_project_dir = require('./utils/change_to_project_dir');
+const NiceError = require('./utils/nice_error.js');
 const interrupt = require('./utils/interrupt');
 const schema = require('./schema');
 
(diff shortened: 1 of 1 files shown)


deontologician:
Just noticed a stray `can to` in the suggestion for disabling security.

cli/src/serve.js
```diff
@@ -244,18 +245,50 @@ const create_insecure_servers = (opts) => {
...
+them in your browser.`
+      );
+      description = `In order to run the server in secure mode (the default), \
+Horizon needs both a certificate file and a key file to encrypt websockets. \
+By default, it looks for horizon-key.pem and horizon-cert.pem \
+files in the current directory.`;
+    } else {
+      // They supplied a cert or key file, so don't give the long
+      // explanation and irrelevant suggestions.
+      suggestions.unshift(`See if the ${type} filename was misspelled`);
```
mlucy: I think this should have a `.` at the end (all the other list items do).

cli/src/utils/nice_error.js
```diff
@@ -0,0 +1,90 @@
...
+        results.push.apply(results, formattedSrc);
+      }
+    }
+    if (this.suggestions) {
+      results.push(
+        '', // extra newline before suggestions
+        chalk.red(
+          this.suggestions.length > 1 ? 'Suggestions:' : 'Suggestion:'));
+      results.push.apply(
+        results, this.suggestions.map((note) => `  ➤ ${note}`));
```
mlucy, line 51: How strongly attached are you to using a Unicode character as the list element prefix?  It struck me as visually weird compared to `-` or `*`, which I'm used to seeing for lists in error messages, and it won't always display correctly for people.

mlucy:
Looks good to me aside from the two minor comments.

cli/src/utils/nice_error.js
deontologician: We use unicode elsewhere, and I tried using *, the problem was with very long suggestions it wasn't visually obvious where one started and the next began

Issue #774, opened 2016-08-22
Server crashes when reconnecting


lirbank:
Server version: 2.0.0-beta-7
Client version: 2.0.0-beta-7

When stopping and starting the server (via cli), to test reconnect (https://github.com/rethinkdb/horizon/issues/772#issuecomment-241564417), I have to try multiple times before the server starts. It seems to be worse when the client is trying (though not sure), eg if I put the retry interval to 10 sec instead of five it seems to go better.

See below how many times I had to launch the server before I got it running:

``` sh
🌄 Horizon ready for connections
^C^C$ hz serve --dev
App available at http://127.0.0.1:8181
RethinkDB
   ├── Admin interface: http://localhost:49532
   └── Drivers can connect to port 49531
Starting Horizon...
error: Unclean shutdown - killing RethinkDB child process
/Users/mikaellirbank/Git/lirbank/horizon/server/src/client.js:33
    if (!this._metadata.is_ready()) {
                       ^

TypeError: Cannot read property 'is_ready' of null
    at new Client (/Users/mikaellirbank/Git/lirbank/horizon/server/src/client.js:33:24)
    at WebSocketServer.websocket.Server.on.on (/Users/mikaellirbank/Git/lirbank/horizon/server/src/server.js:113:39)
    at emitOne (events.js:96:13)
    at WebSocketServer.emit (events.js:188:7)
    at /Users/mikaellirbank/Git/lirbank/horizon/server/node_modules/ws/lib/WebSocketServer.js:91:14
    at completeHybiUpgrade2 (/Users/mikaellirbank/Git/lirbank/horizon/server/node_modules/ws/lib/WebSocketServer.js:284:5)
    at /Users/mikaellirbank/Git/lirbank/horizon/server/node_modules/ws/lib/WebSocketServer.js:297:16
    at Object.accept_protocol (/Users/mikaellirbank/Git/lirbank/horizon/server/src/server.js:36:5)
    at completeHybiUpgrade1 (/Users/mikaellirbank/Git/lirbank/horizon/server/node_modules/ws/lib/WebSocketServer.js:294:32)
    at WebSocketServer.handleHybiUpgrade (/Users/mikaellirbank/Git/lirbank/horizon/server/node_modules/ws/lib/WebSocketServer.js:337:3)
    at WebSocketServer.handleUpgrade (/Users/mikaellirbank/Git/lirbank/horizon/server/node_modules/ws/lib/WebSocketServer.js:173:26)
[...]

PR #778, opened 2016-08-24
Fix crash on client connect when server isn't ready

Tryneus:
Fixes issue #774.  Also fixes a problem I noticed in testing where the primary index is not registered correctly, which resulted in a secondary index being created for just 'id'.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/778)


deontologician:
:lgtm:

Issue #781, opened 2016-08-24
Can't move config.toml


tvld:
I've tried to move config.toml from .hz to project folder, but it does not work even though `--config config.toml`  seems correct...?
Using the dockerfile.yml

deontologician:
Just an update, we closed this since we removed the --config option altogether. There are now several configuration files, and it's better to just have it look in the default location, rather than adding a new flag for each config file.

tvld:
thanks for the update...  ))

PR #788, opened 2016-08-25
Remove --config option from cli commands

Tryneus:
This was a little outdated (we had a way to specify `config.toml`, but not `secrets.toml`, and the help text was a little broken), and didn't work according to #781.  Rather than complicate things by adding even more options, `config.toml` and `secrets.toml` are now always expected to be in `${project_dir}/.hz/`.  This should simplify things a bit.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/788)


deontologician:
:lgtm:

Issue #508, opened 2016-05-25
redirect_uri is hardcoded for https


Penagwin:
This affects [github.js](https://github.com/rethinkdb/horizon/blob/061f784fa6f8cd6b30ffcaf6b88e823e0e386b3f/server/src/auth/github.js) however it does/may affect other Oauth endpoints. Currently the code adds `&redirect_uri=https%3A%2F%2F[Rest of url]` to the url that the user is directed to at github. When this redirect_url isn't the same as the one specified in github's settings it will cause an error, and the github will redirect you back to your app with the error:
`horizon_error=The%20redirect_uri%20MUST%20match%20the%20registered%20callback%20URL%20for%20this%20application.`

This causes an issue in development when working with local addresses where you need to use http (for example with a local address). I suggest that we change [line 108 in utils.js](https://github.com/rethinkdb/horizon/blob/07dc43e7b2a6e869e332be7b7c6b314db7d95ad1/server/src/auth/utils.js#L108) to use the page's current protocol, although I'm not sure how you would want to go about doing that.

My other suggestion is to change [line 28 in github.js](https://github.com/rethinkdb/horizon/blob/next/server/src/auth/github.js#L28) to not include the redirect_url option, as it is optional to use with github, and it is configurable on github's website. Again this may or may not apply to other oath providers, I'm not super familiar with Oauth.

Now I made this change locally, however after logging in with github I get sent to `http://localhost:8181/?horizon_error=session%20expired`, I'm not sure if this is related.

Tryneus:
I may be wrong, but the documentation for most (if not all) auth providers say that you should only use HTTPS for the redirect URI.  At least some of them will change `http` to `https` or reject your redirect URI if it is not secure.  Out of security concerns, I'm not convinced we should allow `http` redirect URIs in the release branch.  Omitting the redirect URI if it is not specified seems fine, as long as the provider handles that correctly.

Keep in mind that it is not too difficult to use HTTPS even when developing on localhost - using a self-signed key, you just need to tell your browser to ignore the warning.

gilbert:
If you want to force `https` during production, great. But forcing it during development is an unnecessary burden. No other library I've encountered has forced me to use `https` for localhost.

Although you say it's "not too difficult", it is still an unnecessary, de-motivational blocker, especially when you're just trying to see if horizon works _at all_ on your own computer.

danielmewes:
[...]

PR #794, opened 2016-08-26
Add option to force https for auth

Tryneus:
Addresses issue #508.  In the CLI, we will set `auth.secure` to the value of `--secure`.  If enabled (the default), we will only redirect back to HTTPS at the end of the oauth flow.  If disabled, we will inspect the request to see if it is on a secure connection (`req.connection.encrypted`).

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/794)

### DIFF
--- a/cli/src/serve.js
+++ b/cli/src/serve.js
@@ -336,6 +336,7 @@ const start_horizon_server = (http_servers, opts) =>
     project_name: opts.project_name,
     access_control_allow_origin: opts.access_control_allow_origin,
     auth: {
+      secure: opts.secure,
       token_secret: opts.token_secret,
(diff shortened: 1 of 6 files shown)


Tryneus:
Thinking about this a bit more, this will make it harder to run `hz serve` behind a reverse proxy - as in that case you would want `--secure=false`, but `auth.secure` to be `true`.  Perhaps we should expose another option on the command-line.

deontologician:
@grey should we close this? your last comment seems to indicate you had second thoughts

Tryneus:
Yeah, I don't think this is sufficient.

Issue #814, opened 2016-09-07
Remove babel-preset-es2015-loose. It has been deprecated.


jimthedev:
For React Native to work I am attempting to get PRs approved upstream in https://github.com/bkonkle/babel-preset-es2015-loose but ultimately it looks like the project has been deprecated. Does anyone know what necessitates the use of this preset? Would it be worth forking it and maintaining it or could we move off to another preset?

PR #815, opened 2016-09-07
Fixes -loose preset for React Native

dalanmiller:
Should handle for #814 but since I know next to nothing about babel @deontologician should look at this.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/815)

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -33,7 +33,7 @@
     "babel-loader": "^6.2.4",
     "babel-plugin-istanbul": "^1.0.3",
     "babel-plugin-transform-runtime": "^6.6.0",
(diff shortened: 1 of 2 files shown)


dalanmiller:
`npm run test` seems to be hanging for me though.

deontologician:
This looks good, we should remove the dependency on loose though

deontologician:
Ok, I got react native working (chat example), and it seems the issue is that we include our .babelrc at all in the package (which we don't need to do). So, moving it out to a .babelrc, removing -loose (just cause) and making sure the .babelrc isn't distributed should work fine

PR #820, opened 2016-09-13
Add support for jsnext:main

simoncarpinter:
- fix typo (`jsnext:main`, not `jsmain:next`)
- include `src/` in distributed version so that `jsnext:main` references a real artifact

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/820)

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -53,7 +53,7 @@
     "webpack": "^1.12.14"
   },
   "main": "lib/index.js",
-  "jsmain:next": "src/index.js",
+  "jsnext:main": "src/index.js",
   "repository": {
     "type": "git",
     "url": "https://github.com/rethinkdb/horizon.git"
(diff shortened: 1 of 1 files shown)


simoncarpinter:
Actually, scratch this.  `jsnext:main` expects to find code that has been transpiled _except_ for its module syntax.  Artifacts in `src/` are not transpiled.

deontologician:
Hey, I think all we need to do is add an es6 output in compile.sh where we use the correct preset to transpile all but imports. Thanks for bringing this to our attention, I'll try getting it to work

simoncarpinter:
Awesome :).  For reference, I realised this was problematic somewhere around 11pm NZT last night and figured I'd close it until I could hack on it a little further -- I meant to come back around to it sometime in the next couple of days after digging into how your compiling/webpack setup works :).  If you're looking into it now then that saves me some digging in unfamiliar territory (I'll admit that JS build tooling is one of my least favourite parts of the ecosystem).

PR #821, opened 2016-09-13
Add Typescript Definitions

marshall007:
In order for Typescript projects to recognize `import`s from npm packages, a `.d.ts` typings definitions file must be present or referenced externally. Adding a `typings` definition allows Horizon to be seamlessly imported into Angular 2 applications.

```
npm install @horizon/client --save
```

``` js
import Horizon from '@horizon/client';
```

Without this, you are forced to either reference Horizon as a global via a `<script>` tag or specify additional `scripts` via `angular-cli.json` and use `require`.

I attempted to cover all of the public API in these definitions, but I'm sure it's not totally complete yet. I think this provides a good enough starting point to at least get things working in Angular 2.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/821)

### DIFF
--- a/client/index.d.ts
+++ b/client/index.d.ts
@@ -1,0 +1,74 @@
+
+import { Observable } from 'rxjs';
+
+declare namespace hz {
+    type Bound = 'open' | 'closed';
+    type Direction = 'ascending' | 'descending';
+    type Primitive = boolean | number | string | Date;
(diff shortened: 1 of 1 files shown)


deontologician:
:lgtm:

somombo:
@deontologician, is this part of the latest release yet?

PR #823, opened 2016-09-14
Fix: move .babelrc to its own file.

### LINKED ISSUE #255: Issues getting Horizon working with React Native
I took horizon for a spin with a react-native project.

I have set up an example project you can use to experiment with it yourselves at [dferber90/horizon-react-native](https://github.com/dferber90/horizon-react-native).
### Current issues

Currently there is at least one issue in [here](https://github.com/rethinkdb/horizon/blob/b5d42cc3b7908fc5bd544725c1ee26d5bb10c665/client/src/auth.js#L71), because `window.location` is undefined in React Native.
After `npm install`ing, it's possible to manually fix this by editing `horizon-react-native/HorizonExampleApp/node_modules/@horizon/client/lib/auth.js` and changing line 98

``` diff
- var parsed = queryParse(window.location.search);
+ var parsed = queryParse(window && window.location && window.location.search);
```

After doing this, press `CMD+R` in the simulator. Now an alert should appear saying the Horizon client is connected... which means it works :)
### Authentication

I haven't fully understood how horizon does authentication yet, so I can describe the problems I discovered only vaguely.
#### Tokens

When using tokens, the token would have to be stored on the device using [AsyncStorage](https://facebook.github.io/react-native/docs/asyncstorage.html#content) instead of `localStorage`. Having low-level functions which allow developers to pass the current token to Horizon from any source and specify how to save any new tokens would be helpful here.
#### OAuth

The current flow of using OAuth is probably not possible with React Native since the OAuth token can't be read form the callback URL, but maybe from the URI Scheme. I've never created OAuth integration for a native device before, so I can't really tell.

### LINKED ISSUE #798: Couldn't find preset "es2015-loose"
Fresh install of version 2. Still can't use horizon with React Native.

Error: Couldn't find preset "es2015-loose" relative to directory "/home/demin/startup/client/node_modules/@horizon/client"

> ```
> at /home/demin/startup/client/node_modules/babel-core/lib/transformation/file/options/option-manager.js:321:17
> ```

Server version: 2
Client version: 2

### LINKED ISSUE #814: Remove babel-preset-es2015-loose. It has been deprecated.

deontologician:
This breaks nothing, our .babelrc shouldn't go out with the distributed package.

One caveat is that if you're doing React Native development, and you have horizon checked out from git and `npm link`ed into your node modules, you'll get the same error as before. (Because the babelrc will be back since it's not the distributed version, just the version from

Fixes #814 
Fixes #798 
Fixes #255 

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/823)

### DIFF
--- a/client/.babelrc
+++ b/client/.babelrc
@@ -1,0 +1,22 @@
+{
+  "presets": [
+    "es2015",
+    {
+      "plugins": [
+        [
+          "transform-runtime",
+          {
+            "polyfill": false
+          }
+        ]
+      ]
+    }
+  ],
+  "env": {
+    "test": {
+      "plugins": [
+        "istanbul"
+      ]
+    }
+  }
+}
--- a/client/package.json
+++ b/client/package.json
@@ -34,7 +34,6 @@
     "babel-plugin-istanbul": "^1.0.3",
     "babel-plugin-transform-runtime": "^6.6.0",
     "babel-preset-es2015": "^6.6.0",
-    "babel-preset-es2015-loose": "^7.0.0",
     "babel-register": "^6.9.0",
     "chai": "^3.5.0",
     "copy-webpack-plugin": "^3.0.1",
@@ -75,28 +74,6 @@
     "dist/horizon-core-dev.js.map",
     "lib/*"
   ],
-  "babel": {
-    "presets": [
-      "es2015-loose",
-      {
-        "plugins": [
-          [
-            "transform-runtime",
-            {
-              "polyfill": false
-            }
-          ]
-        ]
-      }
-    ],
-    "env": {
-      "test": {
-        "plugins": [
-          "istanbul"
-        ]
-      }
-    }
-  },
   "nyc": {
     "all": true,
     "statements": 82.02,
--- a/client/webpack.horizon.config.js
+++ b/client/webpack.horizon.config.js
@@ -73,7 +73,7 @@ module.exports = function(buildTarget) {
           loader: 'babel-loader',
           query: {
             cacheDirectory: true,
-            extends: path.resolve(__dirname, 'package.json'),
+            extends: path.resolve(__dirname, '.babelrc'),
           },
         },
       ],


client/.babelrc
```diff
@@ -0,0 +1,22 @@
...
+            "polyfill": false
+          }
+        ]
+      ]
+    }
+  ],
+  "env": {
+    "test": {
+      "plugins": [
+        "istanbul"
```
dalanmiller: This is quoted beautifully

client/webpack.horizon.config.js
```diff
@@ -73,7 +73,7 @@ module.exports = function(buildTarget) {
           loader: 'babel-loader',
           query: {
             cacheDirectory: true,
-            extends: path.resolve(__dirname, 'package.json'),
```
deontologician: This is bs

client/package.json
```diff
@@ -34,7 +34,6 @@
     "babel-plugin-istanbul": "^1.0.3",
     "babel-plugin-transform-runtime": "^6.6.0",
     "babel-preset-es2015": "^6.6.0",
-    "babel-preset-es2015-loose": "^7.0.0",
```
dalanmiller: Awesome deletion @deontologician !

client/webpack.horizon.config.js
```diff
@@ -73,7 +73,7 @@ module.exports = function(buildTarget) {
...
+            extends: path.resolve(__dirname, '.babelrc'),
```
dalanmiller: `__dirname` is such a hack.

client/package.json
dalanmiller: But. Wait. Don't we need this?

client/webpack.horizon.config.js
deontologician: I dissagree

client/.babelrc
```diff
@@ -0,0 +1,22 @@
+{
+  "presets": [
+    "es2015",
+    {
+      "plugins": [
```
dalanmiller: This is required

deontologician: Ok this is fixed

dalanmiller (approved):

dalanmiller:
Apologies to all for these ridiculous reviews and comments.

hendrikswan:
I still haven't managed to get this working. I went as far as to clone the repo, checking out this branch, publishing an npm package to ensure that it follows the same build process, then installing that npm package into my react native project. 

I am getting a different issue, it's not happy with the "ws" package. I then installed the ws package directly into my project, but then I hit a wall with peer dependencies of "ws", specifically crypto. 

This doesn't feel right.. Any help would be appreciated!

hendrikswan:
Dang! My issue was due to porting some code that I wrote and ran in node directly to my React Native app. In Node, you need to add a global WebSocket variable to make it available to the horizon client code. And I copied that over to my RN app, not realizing that RN also provided a global WebSocket!! 

Mine is working now.

PR #856, opened 2016-10-30
Add AngularJS TodoMVC example

endetti:
A basic example of using [AngularJS](http://angularjs.org/) and [Horizon](http://horizon.io/) to create real-time TodoMVC app.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/856)


marshall007:
@endetti thanks for the PR, this looks great!

The idea with these demos is to provide minimal, comparable examples across frameworks (despite what best practices may be). So in the spirit of our other examples, I think it would make sense to use the `Horizon` client directly in `TodoCtrl` rather than wrapping it in the `TodoStorage` service.

endetti:
@marshall007 I changed the code to fit into the spirit of other examples. Give it a peek! ;)

marshall007:
LGTM - thanks again @endetti!

PR #760, opened 2016-08-18
Plugins

deontologician:
Made this PR for watching plugins grow as a spectator sport.

---

This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/760)

PR #865, opened 2017-01-26
Plugins

Tryneus:
This is a work in progress, but eclipses #760 - this is a fork from that branch to my personal repo.

Several big things remain to do:

* Replace the CLI with one based on Express
* Integrate auth plugins
* Lots of tests
* A complete changelog

The Horizon server has been completely revamped, so any users who directly embedded the server will need non-trivial code changes.  The client is relatively unchanged, it was a goal to make it as painless as possible of an upgrade in the frontend.  Still, a few backwards-incompatibilities surfaced:

* argument validation is now performed server-side, so errors are asynchronous
* batch write errors are no longer an instance of `Error`, but an `{error: ''}` object
* there are no more ordering constraints for chaining terms
* no `state: synced` documents
* the `unconnected` initial horizon state is gone, replaced by `disconnected`
* fewer restrictions on `above`/`below`/`order` - you can now specify the same key in `findAll` without an error (I doubt there's much of a use case for this)
* collections cannot start with 'hz_'

---
This change is [Reviewable](https://reviewable.io/reviews/rethinkdb/horizon/865)


lgleim:
@Tryneus Are you still commited to completing this refactoring process? Either way, could you please give a quick heads up re status so this work can be picked up by someone else otherwise?

sp90:
Whats the status on this?

madwill:
Hi everyone, i'm interested in trying this version of horizon.  I have forked @Tryneus repos and build the grey_plugins branch which compile ok. 

I understand it seems to be using CircleCi for continuous integration. I would be interested in some pointers on how to build and publish the package including client/server together. 

If its somewhat complicated i guess we could publish client and server separately. 

Thank you very much for your time

William

